PR #1066, opened 2015-04-16
Add ProxyFetch::CancelOutstanding and FollowFlushes option

oschaaf:
- Adds ProxyFetch::CancelOutstanding() which helps ngx_pagespeed
  with shutting down quickly by calling Done(false) on all
  live ProxyFetch instances.
- Adds follow_flushes option, which makes ProxyFetch forward
  Flush calls to the RewriteDriver when set.

These changes are needed for: https://github.com/pagespeed/ngx_pagespeed/pull/936

### DIFF
--- a/net/instaweb/rewriter/rewrite_options.cc
+++ b/net/instaweb/rewriter/rewrite_options.cc
@@ -136,6 +136,7 @@ const char RewriteOptions::kFinderPropertiesCacheRefreshTimeMs[] =
     "FinderPropertiesCacheRefreshTimeMs";
 const char RewriteOptions::kFlushBufferLimitBytes[] = "FlushBufferLimitBytes";
 const char RewriteOptions::kFlushHtml[] = "FlushHtml";
+const char RewriteOptions::kFollowFlushes[] = "FollowFlushes";
 const char RewriteOptions::kFlushMoreResourcesEarlyIfTimePermits[] =
     "FlushMoreResourcesEarlyIfTimePermits";
 const char RewriteOptions::kForbidAllDisabledFilters[] =
@@ -1577,6 +1578,10 @@ void RewriteOptions::AddProperties() {
       false, &RewriteOptions::flush_html_, "fh", kFlushHtml,
       kDirectoryScope,
       NULL, true);  // TODO(jmarantz): implement for mod_pagespeed.
+  AddBaseProperty(
+      false, &RewriteOptions::follow_flushes_, "ff", kFollowFlushes,
+      kDirectoryScope,
+      NULL, false);
   AddBaseProperty(
       false, &RewriteOptions::css_preserve_urls_, "cpu",
       kCssPreserveURLs,
--- a/net/instaweb/rewriter/server_context.cc
+++ b/net/instaweb/rewriter/server_context.cc
@@ -884,9 +884,33 @@ void ServerContext::ShutDownDrivers() {
     if (RunningOnValgrind()) {
       timeout_ms *= 20;
     }
+
+    // It is possible that there's still a RewriteContext associated which has
+    // a call scheduled to run, which will drop the last reference and bring
+    // the reference count to 0. That will cause SignalIfRequired to DCHECK if
+    // there is a pending BoundedWaitFor on the driver. To avoid that, add a
+    // user-reference here to pin the driver. Note that in this case, there will
+    // be no user-references left on the driver originally.
+    active->AddUserReference();
     active->BoundedWaitFor(RewriteDriver::kWaitForShutDown, timeout_ms);
     active->Cleanup();  // Note: only cleans up if the rewrites are complete.
     // TODO(jmarantz): rename RewriteDriver::Cleanup to CleanupIfDone.
+  }
+
+  // It's possible that during the BoundedWaitFor the last reference was dropped
+  // in which case the active driver should now be contained in the deferred set.
+  // If it isn't, we need to call Cleanup here. We iterate again here, because
+  // in the earlier iteration other threads may be finishing up drivers and thus
+  // accessing the deferred set.
+  for (RewriteDriverSet::iterator i = active_rewrite_drivers_.begin();
+       i != active_rewrite_drivers_.end(); ++i) {
+    RewriteDriver* driver = *i;
+    if (deferred_release_rewrite_drivers_.find(driver)
+        == deferred_release_rewrite_drivers_.end()) {
+      driver->Cleanup();
+      DCHECK(deferred_release_rewrite_drivers_.find(driver)
+             != deferred_release_rewrite_drivers_.end());
+    }
   }
 }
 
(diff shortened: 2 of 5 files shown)


pagespeed/automatic/Makefile
```diff
@@ -24,7 +24,7 @@
 #
 # When running this Makefile from the 'automatic' directory then it will
 # be set automatically.
-MOD_PAGESPEED_ROOT = $(shell cd ../../..; pwd)
+MOD_PAGESPEED_ROOT = $(shell cd ../..; pwd)
```
oschaaf: @jeffkaufman I just noticed I left this in, but I'm not sure this is still needed?

jeffkaufman: Something is still needed, but I think not quite this.  Paths are slightly different inside Google and I'm pretty sure we need ../../.. outside Google and ../.. inside.  I'll fix this in a separate change.

oschaaf: @jeffkaufman Ack - removed from this change.

jeffkaufman:
@morlovich Could you review this change, and the companion change pagespeed/ngx_pagespeed#936 ?

net/instaweb/rewriter/rewrite_options.cc
```diff
@@ -1578,6 +1579,10 @@ void RewriteOptions::AddProperties() {
       kDirectoryScope,
       NULL, true);  // TODO(jmarantz): implement for mod_pagespeed.
   AddBaseProperty(
+      false, &RewriteOptions::follow_flushes_, "ff", kFollowFlushes,
```
morlovich, line 1582: Are you sure you don't just want to set the default of FlushHtml to true?

pagespeed/automatic/proxy_fetch.cc
```diff
@@ -77,7 +77,8 @@ ProxyFetchFactory::ProxyFetchFactory(ServerContext* server_context)
       timer_(server_context->timer()),
       handler_(server_context->message_handler()),
       outstanding_proxy_fetches_mutex_(
-          server_context->thread_system()->NewMutex()) {
+      server_context->thread_system()->NewMutex()),
```
morlovich, line 80: The +4 indent is needed per style guide...

net/instaweb/rewriter/server_context.cc
```diff
@@ -884,10 +884,34 @@ void ServerContext::ShutDownDrivers() {
...
+    // there is a pending BoundedWaitFor on the driver. To avoid that, add a
+    // user-reference here to pin the driver. Note that in this case, there will
+    // be no user-references left on the driver originally.
+    active->AddUserReference();
     active->BoundedWaitFor(RewriteDriver::kWaitForShutDown, timeout_ms);
     active->Cleanup();  // Note: only cleans up if the rewrites are complete.
     // TODO(jmarantz): rename RewriteDriver::Cleanup to CleanupIfDone.
   }
+
+  // It's possible that during the BoundedWaitFor the last reference was dropped
```
morlovich, line 900: How does that happen with the AddUserReference?

pagespeed/automatic/proxy_fetch.cc
```diff
@@ -90,6 +91,37 @@ ProxyFetchFactory::~ProxyFetchFactory() {
...
+
+  // First wait any current scheduled CompleteFinishParse calls to round up,
+  // so we don't have to worry about races w/regard to done_outstanding_ and
+  // finishing_, avoiding the need to take a lock on proxy_fetch.
+  while (proxy_fetches_done_in_flight_.value() != 0) {
+    timer_->SleepUs(sleep_us);
+  }
+
+  // Any outstanding fetches left do not have a Done() call initiated on them.
+  // So we don't have to worry about done_outstanding_ and finishing_
```
morlovich, line 105: What's preventing them from being stuck right at the { of Done?

pagespeed/automatic/proxy_fetch.cc
```diff
@@ -90,6 +91,37 @@ ProxyFetchFactory::~ProxyFetchFactory() {
...
+      fetch = *outstanding_proxy_fetches_.begin();
+      outstanding_proxy_fetches_.erase(fetch);
+    }
+    fetch->Done(false);
+  }
+
+  // Wait for any deferred finalization to round up before exiting, releasing
+  // any associated drivers. 
+  while (proxy_fetches_done_in_flight_.value() != 0) {
+    timer_->SleepUs(sleep_us);  }
```
morlovich, line 122: line break after ;

pagespeed/automatic/proxy_fetch.cc
```diff
@@ -90,6 +91,37 @@ ProxyFetchFactory::~ProxyFetchFactory() {
...
+    ProxyFetch* fetch;
+    {
+      ScopedMutex lock(outstanding_proxy_fetches_mutex_.get());
+      if (outstanding_proxy_fetches_.empty()) {
+        break;
+      }
+      fetch = *outstanding_proxy_fetches_.begin();
+      outstanding_proxy_fetches_.erase(fetch);
+    }
+    fetch->Done(false);
```
morlovich, line 116: I don't really see anything that's preventing someone else from calling Done on the fetch... or have we shut down the threads feeding it data at this point?

net/instaweb/rewriter/rewrite_options.cc
oschaaf, line 1582: The reasons for defaulting this flag to false:
- For very chunky html responses a default of true could hurt
- I understood from @jmarantz that the code that hides under FlushHtml is pretty sensitive. 

What do  you think?

net/instaweb/rewriter/server_context.cc
oschaaf, line 900: @morlovich I'm not sure I understand your question 100%.

A few lines above, there's this code:

```
+    active->AddUserReference();
     active->BoundedWaitFor(RewriteDriver::kWaitForShutDown, timeout_ms);            active->BoundedWaitFor(RewriteDriver::kWaitForShutDown, timeout_ms);
     active->Cleanup();  // Note: only cleans up if the rewrites are complete.           active->Cleanup();  // Note: only cleans up if the rewrites are complete.

```

Does that answer your question?

pagespeed/automatic/proxy_fetch.cc
oschaaf, line 116: After a server calls `CancelOutstanding()`, no other `ProxyFetch` instances may be created  - and existing ones should not be touched.

I see that this is not sufficiently documented in the header, and it might make sense to DCHECK when someone attempts this..

jeffkaufman:
where are we with this?

jeffkaufman:
Looks like this was updated and re-submitted as https://github.com/pagespeed/mod_pagespeed/issues/1332 , and got merged as 02de03e

PR #1140, opened 2015-09-09
take CXX from environment variable

### LINKED ISSUE #1131: g++ location hardcoded on makefile for psol .a
https://github.com/pagespeed/mod_pagespeed/blob/master/pagespeed/automatic/Makefile#L222

should take the CXX environment variable.

crowell:
stop hardcoding /usr/bin/g++, instead just use CXX environment var.

closes #1131


crowell:
this is required for building with gcc in a different location, such as with the redhat developer toolset on centos 5.

PR #1150, opened 2015-10-07
fix build of boringssl with newer compilers (gcc5)

crowell:
getaddrinfo is a POSIX extension.

newer compilers (gcc5?) require additional cflags to use getaddrinfo and
friends.


oschaaf:
FWIW, LGTM!

jeffkaufman:
LGTM

PR #1151, opened 2015-10-07
Css flattening: recompute flattening_failed_ on cache hit

oschaaf:
Fixes https://github.com/pagespeed/mod_pagespeed/issues/1092

### DIFF
--- a/net/instaweb/rewriter/css_flatten_imports_test.cc
+++ b/net/instaweb/rewriter/css_flatten_imports_test.cc
@@ -46,6 +46,7 @@ const char kTwoLevelsDownFile1[] = "assets/nested/nested1.css";
 const char kTwoLevelsDownFile2[] = "assets/nested/nested2.css";
 const char k404CssFile[] = "404.css";
 const char kSimpleCssFile[] = "simple.css";
+const char kComplexCssFile[] = "complex.css";
 
 // Contents of resource files. Already minimized. NOTE relative paths!
 static const char kTwoLevelsDownContents1[] =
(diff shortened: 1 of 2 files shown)


jeffkaufman:
I'll merge this in a bit; thanks for the reminder.

I need to pull this into the google-internal copy of the repo as well.

morlovich:
Render really shouldn't be doing computation (like re-parsing the entire CSS), the efficient way of doing that would be to store the bit in CachedResult proto and restore it.

jmaessen:
Maks points out that we've already paid the parsing cost here, and that this effectively re-parses the CSS.  If that's true, we should just save the value of the bit when we first parse (by adding a bit to the proto) and using that here instead.

jeffkaufman:
We hit this rarely, and it's in the optimization phase not the serving phase.  I'm happy just re-parsing the CSS in this case and moving on.

oschaaf:
With this pull I was hoping that the benefit of less code would outweigh the performance cost for re-parsing the CSS upon re-optimization. (also see the last bit of the discussion on https://github.com/pagespeed/mod_pagespeed/issues/1092)

morlovich:
Sorry, you are probably right re: complexity of caching tradeoff, especially since CssHiearchy has a lot of state beyond just that bit, and as this is a nested rewrite the top-level caching should help most of the time. In general, of course, Render() methods ought to be entirely based on CachedResult in that they simply do not have the input available in some cases (since we don't fetch it on a cache hit), but here it's saved inside CachedResult anyway. This probably means that CssHiearchy ought not have existed in the first place, but this sure as heck isn't a reason or place to change that....

I also have to apologize for my slowness on this --- I was being overly indecisive in picking between two imperfect options.

jeffkaufman:
merged as c5d13af

PR #1157, opened 2015-10-23
Switch to getting libpng from its git, and update to 1.2.52

morlovich:
(Which tweaks one of its headers to be C++-11 compilable)


jeffkaufman:
LGTM!

PR #1161, opened 2015-10-27
nps-1015: IPRO+MPD support in ngx_pagespeed

oschaaf:
- Move the Apache tests for IPRO + MPD to system/system_test.sh
- Add a flag 'trusted_input_' in ProxyFetch to allow ngx_pagespeed
  to transform html but not proxy external html fetched via MPD.

MPS-side of the fix for:
https://github.com/pagespeed/ngx_pagespeed/issues/1015

### DIFF
--- a/pagespeed/automatic/proxy_fetch.h
+++ b/pagespeed/automatic/proxy_fetch.h
@@ -342,7 +342,7 @@ class ProxyFetch : public SharedAsyncFetch {
   //
   // This is used only for testing.
   static const int kTestSignalTimeoutMs = 200;
-
+  void set_trusted_input(bool trusted_input) { trusted_input_ = trusted_input; }
  protected:
   // protected interface from AsyncFetch.
   virtual void HandleHeadersComplete();
@@ -513,6 +513,10 @@ class ProxyFetch : public SharedAsyncFetch {
 
   // Set to true if this proxy_fetch is the result of a distributed fetch.
   bool distributed_fetch_;
+  
+  // Set to true if this proxy_fetch is actually operating on trusted
+  // (non-proxied) content.
+  bool trusted_input_;
 
   DISALLOW_COPY_AND_ASSIGN(ProxyFetch);
 };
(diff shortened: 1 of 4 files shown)


pagespeed/automatic/proxy_fetch.h
```diff
@@ -342,7 +342,7 @@ class ProxyFetch : public SharedAsyncFetch {
   //
   // This is used only for testing.
   static const int kTestSignalTimeoutMs = 200;
-
+  void SetTrustedInput() { trusted_input_ = true; }
```
jmarantz: nit: style guide says to spell this 'set_trusted_input'.

oschaaf: @jmarantz  re: "nit: style guide says to spell this 'set_trusted_input'."
renamed accordingly

jmarantz: Sorry... One more thing....And it should take the new value as a formal.

oschaaf: @jmarantz re: "And it should take the new value as a formal."
done

pagespeed/automatic/proxy_fetch.h
```diff
@@ -513,6 +513,8 @@ class ProxyFetch : public SharedAsyncFetch {
 
   // Set to true if this proxy_fetch is the result of a distributed fetch.
   bool distributed_fetch_;
+  
+  bool trusted_input_;
```
jmarantz, line 519: add comment here explaining what this is used for.

jmarantz, line 519: otherwise, lgtm

jeffkaufman, line 519: Add a comment like:

```
  // Set to true if this proxy_fetch is actually operating on trusted            
  // (non-proxied) content.
```

oschaaf:
@jmarantz @jeffkaufman suggested comment for trusted_input_  added

jeffkaufman:
LGTM.  I'll merge this and the nginx side once I finish running our internal tests.

jeffkaufman:
Internal tests pass, merging.

PR #1168, opened 2015-11-02
Add newlines when missing at ends of combined CSS files,

morlovich:
in order to be more robust against problems involving unmatched
quotes.

See https://github.com/pagespeed/mod_pagespeed/issues/1165

### DIFF
--- a/net/instaweb/rewriter/css_combine_filter.cc
+++ b/net/instaweb/rewriter/css_combine_filter.cc
@@ -157,7 +157,7 @@ class CssCombineFilter::CssCombiner : public ResourceCombiner {
     return &kContentTypeCss;
   }
 
-  virtual bool WritePiece(int index, const Resource* input,
+  virtual bool WritePiece(int index, int num_pieces, const Resource* input,
                           OutputResource* combination, Writer* writer,
                           MessageHandler* handler);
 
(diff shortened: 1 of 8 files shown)


net/instaweb/rewriter/css_combine_filter.cc
```diff
@@ -504,6 +504,11 @@ bool CssCombineFilter::CssCombiner::WritePiece(
       input_url, combination->resolved_base(), contents, writer, handler)) {
     case RewriteDriver::kNoResolutionNeeded:
       ret = writer->Write(contents, handler);
+      if (!contents.ends_with("\n") && ret) {
+        // Ensure that we add a new line at the end, to make sure that any
+        // unopened strings get closed in a predictable manner.
+        ret = writer->Write("\n", handler);
+      }
```
jmaessen, line 512: I notice (from the tests) you seem to be adding the trailing newline unconditionally at the end of a CSS block.  What would it take to just insert it in the between pairs of CSS and / or avoid the trailing newline?

morlovich:
Just changing the signature to the method to pass the total count, and adjusting the JS combiner to new sig. Also finding the instruction on how to update the CLs here, I suppose :)

jeffkaufman:
@morlovich 

> Just changing the signature to the method to pass the total count, and adjusting the JS combiner to new sig. 

That seems worth it, no?

> Also finding the instruction on how to update the CLs here, I suppose :)

Export again, force-push.  If you're having trouble come poke me.

morlovich:
> Export again, force-push. If you're having trouble come poke me.
> One concern for that: I needed to use --upstream=origin (or was it --set-upstream?) on the original push. I am not at all sure as to what would happen with a force push.

jeffkaufman:
@morlovich 

> I am not at all sure as to what would happen with a force push.

That's not a problem.  Force-pushing your branch will overwrite only your branch (with the updated version).

With our project it's safe to force-push any branch starting with your username.

morlovich:
Hopefully updated to not add \n after last one.

jeffkaufman:
LGTM

jmaessen:
LGTM

PR #1177, opened 2015-11-10
Honor disallow in defer_js

morlovich:
See: https://github.com/pagespeed/mod_pagespeed/issues/1173


jeffkaufman:
LGTM

jmaessen:
So...  How do I indicate to mod_pagespeed that I don't want it to rewrite my javascript, but I do want it deferred?

PR #1178, opened 2015-11-10
insert_ga: support content experiments, support analytics.js

jeffkaufman:
Previously insert_ga only supported old-style ga.js, and when used with
experiments could only report to a GA custom variable.  This change lets
insert_ga add the new-style analytics.js snippet, report to a GA content
experiment, and supports modifying both old= and new-style snippets.

Nginx side of the change: https://github.com/pagespeed/ngx_pagespeed/pull/1042

Fixes https://github.com/pagespeed/mod_pagespeed/issues/873


net/instaweb/rewriter/public/insert_ga_filter.h
```diff
@@ -34,14 +34,28 @@ class RewriteDriver;
...
+extern const char kContentExperimentsJsClientUrl[];
+extern const char kContentExperimentsSetChosenVariantSnippet[];
+extern const char kContentExperimentsSetExpAndVariantSnippet[];
 extern const char kGASpeedTracking[];
 
-// This class is the implementation of insert_ga_snippet filter, which adds
-// a Google Analytics snippet into the head of html pages.
+// This class is the implementation of the insert_ga filter, which handles:
+// * Adding a Google Analytics snippet to html pages.
+// * Adding js to reporting experiment data to Google Analytics.
```
morlovich: EGRAMMAR

net/instaweb/rewriter/insert_ga_filter.cc
```diff
@@ -141,56 +199,82 @@ void InsertGAFilter::StartElementImpl(HtmlElement* element) {
...
+  DCHECK(text.empty() != url.empty()) <<
+      "Exactly one of text/url should be set.";
+
+  HtmlElement* parent;
+  if (append_child) {
+    parent = current_element;
+  } else {
+    parent = current_element->parent();
+    if (parent == NULL) {
+      LOG(INFO) << "Null parent in insert_ga can't insert following node.";
```
oschaaf, line 247: Should this be a warning?

jeffkaufman:
updated and force-pushed

net/instaweb/rewriter/public/insert_ga_filter.h
jeffkaufman: fixed

net/instaweb/rewriter/insert_ga_filter.cc
jeffkaufman, line 247: This could happen outside of the operator's control if pagespeed was running in a proxy or something, which makes me wary about calling it a warning.

oschaaf, line 247: Ah, yes, INFO makes sense to me now.

jeffkaufman:
This was merged as 7c38c9076f86e0d448720c7076abe030db9f7dd5

PR #1195, opened 2015-11-29
headers sanitization: additional hop-by-hop header handling

### LINKED ISSUE #1090: MapProxyDomain vs Response with "Alternate-Protocol" header
Noticed at iispeed.com
When applying MapProxyDomain to `https://www.google-analytics.com/analytics.js`:
This response header will get forwarded: `Alternate-Protocol: 443:quic,p=1` in the optimized response.
That seems optimistic, I wish our server supported that :-)

Note that there's a hack in place to supress this right now, so it won't reproduce.

oschaaf:
Details: 
- Adds `Connection:` to the list of headers that contain fields separated by ","
- Marks `Alt-Svc` and `Alternate-Protocol` as hop-by-hop, so we will treat them as such.
- Adds sanitization of headers marked as hop-by-hop in `Connection:` headers as per rfc.
- Moves initialization of 'AtExitManager' out of the css filter to the more central `ProcessContext` as this change adds another dependancy on it -- and we must initialize it exactly once.
- Changes the header definitions to be lazy initialized to avoid extra allocations when manipulating headers, while avoiding static initialization. 

Fixes #1090

### DIFF
--- a/net/instaweb/rewriter/css_filter.cc
+++ b/net/instaweb/rewriter/css_filter.cc
@@ -73,13 +73,6 @@
 #include "pagespeed/opt/logging/enums.pb.h"
 #include "webutil/css/parser.h"
 
-#include "base/at_exit.h"
-
-namespace {
-
-base::AtExitManager* at_exit_manager = NULL;
-
-}  // namespace
 
 namespace net_instaweb {
 
@@ -894,8 +887,6 @@ T* MergeArrays(const T* a, int a_size, const T* b, int b_size, int* out_size) {
 }  // namespace
 
 void CssFilter::Initialize() {
-  InitializeAtExitManager();
-
   CHECK(merged_filters_ == NULL);
 #ifndef NDEBUG
   for (int i = 1; i < kRelatedFiltersSize; ++i) {
@@ -917,12 +908,6 @@ void CssFilter::Initialize() {
 }
 
 void CssFilter::Terminate() {
-  // Note: This is not thread-safe, but I don't believe we need it to be.
-  if (at_exit_manager != NULL) {
-    delete at_exit_manager;
-    at_exit_manager = NULL;
-  }
-
   CHECK(merged_filters_ != NULL);
   delete [] merged_filters_;
   merged_filters_ = NULL;
@@ -934,13 +919,6 @@ void CssFilter::Terminate() {
 void CssFilter::AddRelatedOptions(StringPieceVector* target) {
   for (int i = 0, n = arraysize(kRelatedOptions); i < n; ++i) {
     target->push_back(kRelatedOptions[i]);
-  }
-}
-
-void CssFilter::InitializeAtExitManager() {
-  // Note: This is not thread-safe, but I don't believe we need it to be.
-  if (at_exit_manager == NULL) {
-    at_exit_manager = new base::AtExitManager;
   }
 }
 
--- a/pagespeed/kernel/http/http_names.cc
+++ b/pagespeed/kernel/http/http_names.cc
@@ -19,6 +19,8 @@
 #include "pagespeed/kernel/http/http_names.h"
 
 
+#include "base/at_exit.h"
+#include "base/lazy_instance.h"
 #include "base/logging.h"
 #include "pagespeed/kernel/base/string_util.h"
 
@@ -33,6 +35,8 @@ const char HttpAttributes::kAccessControlAllowCredentials[] =
     "Access-Control-Allow-Credentials";
 const char HttpAttributes::kAge[] = "Age";
 const char HttpAttributes::kAllow[] = "Allow";
+const char HttpAttributes::kAltSvc[] = "Alt-Svc";
+const char HttpAttributes::kAlternateProtocol[] = "Alternate-Protocol";
 const char HttpAttributes::kAttachment[] = "attachment";
 const char HttpAttributes::kAuthorization[] = "Authorization";
 const char HttpAttributes::kCacheControl[] = "Cache-Control";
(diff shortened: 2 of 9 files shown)


net/instaweb/rewriter/css_filter.cc
```diff
@@ -73,13 +73,6 @@
 #include "pagespeed/opt/logging/enums.pb.h"
 #include "webutil/css/parser.h"
 
-#include "base/at_exit.h"
-
-namespace {
-
-base::AtExitManager* at_exit_manager = NULL;
```
morlovich, line 80: I can't say I am a fan of this stuff, but that seems unrelated to this change?

oschaaf, line 80: @morlovich Sorry, late reply. 
Do you mean to ask why this is deleted here? I had to move this initialization over to a more central location because it has to be performed exactly once. Its new home ended up being `process_context.cc` (see later on in this change). 
Also see the changes in `http_names.cc`, which add another dependency by using`LazyInstance`.

pagespeed/kernel/http/response_headers_test.cc
```diff
@@ -2104,4 +2113,32 @@ TEST_F(ResponseHeadersTest, ClearOptionCookies) {
...
+  headers.Add(HttpAttributes::kTransferEncoding, "foo");
+  headers.Add(HttpAttributes::kUpgrade, "foo");
+  headers.Add(HttpAttributes::kAlternateProtocol, "foo");
+  headers.Add(HttpAttributes::kCacheControl, "foo");
+  // foo: foo is be referenced in "Connection: Foo", and therefore is marked
+  // as hop-by-hop and as such candidate for sanitization.
+  headers.Add("foo", "foo");
+
+  // Before sanitization, we should have headers.
+  EXPECT_NE("HTTP/1.0 0 (null)\r\n\r\n", headers.ToString());
```
jeffkaufman: This seems kind of specific for an `EXPECT_NE`?

pagespeed/kernel/http/response_headers.cc
```diff
@@ -491,17 +491,85 @@ void ResponseHeaders::SetOriginalContentLength(int64 content_length) {
...
+      const StringPieceVector& end_to_end = HttpAttributes::SortedEndToEndHeaders();
+      StringCompareInsensitive compare;
+      StringPieceVector split;
+      SplitStringPieceToVector(headers->Get(i).value(), ",", &split, true);
+      if (split.empty()) {
+        split.push_back(headers->Get(i).value());
+      }
+
+      // Check each value in Connection: val1, val2, ...
+      for (int j = 0; j < (int)split.size(); ++j) {
```
jeffkaufman: nit: normally we write this like the loop above, with `i=0, n=size...`

pagespeed/kernel/http/response_headers.cc
```diff
@@ -491,17 +491,85 @@ void ResponseHeaders::SetOriginalContentLength(int64 content_length) {
...
+
+      changed = RemoveAll(*v[i]) || changed;
+    }
+  }
+
+  // Remove cookies plus any well-known hop-by-hop headers, which we will never
+  // store in a cache.
+  const StringPieceVector& names_to_sanitize = HttpAttributes::SortedHopByHopHeaders();
+  changed = RemoveAllFromSortedArray(&names_to_sanitize[0],
+                                  names_to_sanitize.size()) || changed;
```
jeffkaufman: nit: indentation

jeffkaufman:
LGTM

@morlovich do you want to look more or are you happy?

pagespeed/kernel/http/response_headers_test.cc
oschaaf: @jeffkaufman re :"This seems kind of specific for an  EXPECT_NE ?"
I removed this test, I don't think it added much value. Is that ok?

oschaaf:
@jeffkaufman comments addressed & force pushed

jeffkaufman:
@morlovich LGTY?

morlovich:
Looking, thanks for the ping.

pagespeed/kernel/http/response_headers.cc
```diff
@@ -539,17 +539,90 @@ void ResponseHeaders::SetOriginalContentLength(int64 content_length) {
...
+        StringPiece val = split[j];
+        TrimWhitespace(&val);
+        // Skip values that are connection-tokens, empty, or are defined
+        // as being end-to-end.
+        if (val.empty() || StringCaseEqual(val, "keep-alive") ||
+            StringCaseEqual(val, "close") ||
+            StringCaseStartsWith(val, "timeout=") ||
+            StringCaseStartsWith(val, "max=") ||
+            std::binary_search(
+              end_to_end.begin(), end_to_end.end(), val, compare)) {
```
morlovich: Could you factor out this condition into a separate function? Seems a bit too big to be repeated.

pagespeed/kernel/http/response_headers.cc
```diff
@@ -539,17 +539,90 @@ void ResponseHeaders::SetOriginalContentLength(int64 content_length) {
...
+          continue;
+        }
+
+        // Find the position at which we should insert to keep the list sorted.
+        std::vector<StringPiece>::iterator up =
+          std::lower_bound(more_names.begin(), more_names.end(), val, compare);
```
morlovich: +2 indent here, and also in argument lists in calls to std::binary_search above.

pagespeed/kernel/http/response_headers.cc
```diff
@@ -539,17 +539,90 @@ void ResponseHeaders::SetOriginalContentLength(int64 content_length) {
...
+
+        // Insert when the entry is not already contained.
+        if (more_names.size() == 0 || up == more_names.end() ||
```
morlovich: Can drop the first clause, I think, FWIW.

pagespeed/kernel/http/http_names.cc
```diff
@@ -189,43 +193,138 @@ const char* HttpStatus::GetReasonPhrase(HttpStatus::Code rc) {
...
+class EndToEndHeadersContainer {
+ public:
+  EndToEndHeadersContainer() {
+    BuildEndToEnd();
+    BuildHopByHop();
+    BuildCachingHeadersToBeRemoved();
+  }
+
+  void BuildEndToEnd() {
+    // See http://www.w3.org/Protocols/rfc2616/rfc2616-sec13.html#sec13.5.1
```
morlovich, line 207: Not worth changing, but: you may want to start using the new HTTP1.1 RFCs (RFC7230-RFC7235), not the obsolete RFC2616, as they should be more precise.

morlovich:
Merged (with some memory usage/deep copy tweaks)

jeffkaufman:
https://github.com/pagespeed/mod_pagespeed/commit/d43f40b3c7b3b27dfce3350cea2d8b62f83cdc27

PR #1196, opened 2015-11-30
convert-meta-tags: don't allow newlines when converting meta tags.

oschaaf:
This change makes ResponseHeaders::MergeContentType reject values
containing unprintable characters.

Fixes https://github.com/pagespeed/mod_pagespeed/issues/1083

### DIFF
--- a/pagespeed/kernel/http/response_headers.cc
+++ b/pagespeed/kernel/http/response_headers.cc
@@ -321,6 +321,12 @@ bool ResponseHeaders::CombineContentTypes(const StringPiece& orig,
 }
 
 bool ResponseHeaders::MergeContentType(const StringPiece& content_type) {
+  for (size_t i = 0; i < content_type.size(); i++) {
+    if (!IsNonControlAscii(content_type[i])) {
+      return false;
+    }
+  }
+
   bool ret = false;
   ConstStringStarVector old_values;
   Lookup(HttpAttributes::kContentType, &old_values);
--- a/net/instaweb/rewriter/meta_tag_filter_test.cc
+++ b/net/instaweb/rewriter/meta_tag_filter_test.cc
@@ -67,6 +67,19 @@ TEST_F(MetaTagFilterTest, TestTags) {
       << *values[0];
 }
 
+const char kMetaTagDocInvalidAttribute[] =
+    "<html><head>"
+    "<meta http-equiv=\"Content-Type\" content=\"text/html; charset=U\r\nTF-8\">"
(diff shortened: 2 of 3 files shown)


pagespeed/kernel/http/response_headers.cc
```diff
@@ -321,23 +321,28 @@ bool ResponseHeaders::CombineContentTypes(const StringPiece& orig,
...
-  if (old_values.size() < 1) {
-    ret = CombineContentTypes("", content_type);
-  } else if (old_values.size() == 1) {
-    StringPiece old_val(*old_values[0]);
-    ret = CombineContentTypes(old_val, content_type);
-  }
+
   if (ret) {
-    cache_fields_dirty_ = true;
+    Lookup(HttpAttributes::kContentType, &old_values);
```
crowell: check return value of `Lookup`, false means lookup failed.

oschaaf: @crowell  This is exactly like it was before this change. Double checking, it looks to me like this should work ok when `Lookup` would fail (`old_values` should be unchanged). Do you have additional handling in mind for the case where `Lookup` returns false?

crowell: I think i was just looking at the diff wrong in github, your change is unrelated to this block, sorry for the noise.

pagespeed/kernel/http/response_headers.cc
```diff
@@ -321,23 +321,28 @@ bool ResponseHeaders::CombineContentTypes(const StringPiece& orig,
 }
 
 bool ResponseHeaders::MergeContentType(const StringPiece& content_type) {
-  bool ret = false;
+  bool ret = std::find_if(
+      content_type.begin(), content_type.end(), iscntrl) == content_type.end();
```
morlovich: I have to admit to a certain paranoia about the likes of iscntrl, due to their locale-sensitivity.
Maybe our inline bool IsNonControlAscii(char c) from string_util.h would be an acceptable substitute?

oschaaf: @morlovich thanks, sounds like a good idea, I'll swap in `IsNonControlAscii()`

oschaaf: @morlovich updated & force pushed

jeffkaufman:
LGTM

I'll merge this.

jeffkaufman:
Merged as 2cea12ce1157ff4e7ae0944578eeeef88110eacf

PR #1200, opened 2015-12-08
New option: AddResourceHeader

oschaaf:
Allows setting headers on PageSpeed resources for CORS support.

Fixes https://github.com/pagespeed/ngx_pagespeed/issues/1005

### DIFF
--- a/net/instaweb/rewriter/public/rewrite_options.h
+++ b/net/instaweb/rewriter/public/rewrite_options.h
@@ -414,6 +414,7 @@ class RewriteOptions {
   static const char kInlineResourcesWithoutExplicitAuthorization[];
   static const char kRetainComment[];
   // 2-argument ones:
+  static const char kAddResourceHeader[];
   static const char kCustomFetchHeader[];
   static const char kLoadFromFile[];
   static const char kLoadFromFileMatch[];
@@ -1041,6 +1042,34 @@ class RewriteOptions {
   }
   void SetRewriteLevel(RewriteLevel level) {
     set_option(level, &level_);
+  }
+
+  // Returns true iff given name and value are valid to set as a http header.
+  // If false is returned, error_message will have a descriptive failure
+  // message set.
+  bool ValidateConfiguredHttpHeader(const GoogleString& name,
+                                    const GoogleString& value,
+                                    GoogleString* error_message);
+
+  // If false was returned, no changes are made, and error_message will be
+  // assigned a failure description. If true was returned, the name/value
+  // passed validation and have been appended to resource_headers_.
+  // Multiple calls to ValidateAndAddResourceHeader with the same name will
+  // result in multiple headers being appended with the same name.
+  bool ValidateAndAddResourceHeader(const StringPiece& name,
+                                    const StringPiece& value,
+                                    GoogleString* error_message);
+
+  // Unconditionally appends the given name / value to resource_headers_. Use
+  // ValidateAndAddResourceHeader if you need validation.
+  void AddResourceHeader(const StringPiece& name, const StringPiece& value);
+
+  const NameValue* resource_header(int i) const {
+    return resource_headers_[i];
+  }
+
+  int num_resource_headers() const {
+    return resource_headers_.size();
   }
 
   // Specify a header to insert when fetching subresources.
@@ -3547,6 +3576,9 @@ class RewriteOptions {
   static void InitFilterIdToEnumArray();
   static void InitOptionIdToPropertyArray();
   static void InitOptionNameToPropertyArray();
+  // Inits fixed_resource_headers_ to a sorted list of headers not allowed in
+  // AddResourceHeader.
+  static void InitFixedResourceHeaders();
 
   // Helper for converting the result of SetOptionFromNameInternal into
   // a status/message pair. The returned result may be adjusted from the passed
@@ -4220,6 +4252,11 @@ class RewriteOptions {
   int experiment_percent_;  // Total traffic going through experiments.
   std::vector<ExperimentSpec*> experiment_specs_;
 
+  // Headers to add resource responses.
+  // TODO(oschaaf): Wrap this and custom_fetch_headers_ in CopyOnWrite through
+  // a new class. Move header validations over there as well so we can share
+  // them.
+  std::vector<NameValue*> resource_headers_;
   // Headers to add to subresource requests.
   std::vector<NameValue*> custom_fetch_headers_;
 
--- a/net/instaweb/rewriter/resource_fetch.cc
+++ b/net/instaweb/rewriter/resource_fetch.cc
@@ -171,6 +171,11 @@ void ResourceFetch::HandleHeadersComplete() {
   DCHECK(!response_headers()->Lookup(HttpAttributes::kSetCookie2, &v));
   response_headers()->RemoveAll(HttpAttributes::kSetCookie);
   response_headers()->RemoveAll(HttpAttributes::kSetCookie2);
+  for (int i = 0; i < driver_->options()->num_resource_headers(); ++i) {
+    const RewriteOptions::NameValue* nv =
+        driver_->options()->resource_header(i);
+    response_headers()->Add(nv->name, nv->value);
+  }
 
   // "Vary: Accept-Encoding" for all resources that are transmitted compressed.
   // Server ought to set these, I suppose.
--- a/net/instaweb/rewriter/resource_fetch_test.cc
+++ b/net/instaweb/rewriter/resource_fetch_test.cc
@@ -43,6 +43,7 @@ namespace {
 
 const char kCssContent[] = "* { display: none; }";
 const char kMinimizedCssContent[] = "*{display:none}";
+const char kValue[] = "Value";
 
 class ResourceFetchTest : public RewriteTestBase {
 };
@@ -66,6 +67,52 @@ TEST_F(ResourceFetchTest, BlockingFetch) {
           CreateRequestContext());
   RewriteOptions* custom_options =
       server_context()->global_options()->Clone();
+
+  GoogleString err;
+  // Tell ResourceFetch to add a few response headers to its results.
+  // Empty field name gets rejected.
+  EXPECT_FALSE(custom_options->ValidateAndAddResourceHeader("", "Bar", &err));
+
+  // Empty field value gets accepted.
+  EXPECT_TRUE(custom_options->ValidateAndAddResourceHeader(
+      "X-Foo-Empty", "", &err));
+
+  // No control characters allowed in field name
+  EXPECT_FALSE(custom_options->ValidateAndAddResourceHeader(
+      "X-Foo\ncontinue", "Bar", &err));
+
+  // No control characters allowed in field value
+  EXPECT_FALSE(custom_options->ValidateAndAddResourceHeader(
+      "X-Foo", "Bar\ncontinue", &err));
+
+  // No separators should be accepted in the field name
+  EXPECT_FALSE(custom_options->ValidateAndAddResourceHeader(
+      "X-Fo;o", "Bar", &err));
+
+  // Hop by hop headers should be refused.
+  EXPECT_FALSE(custom_options->ValidateAndAddResourceHeader(
+      "Connection", "close", &err));
+
+  // Cache-control header should be refused.
+  EXPECT_FALSE(custom_options->ValidateAndAddResourceHeader(
+      "Cache-Control", "private", &err));
+
+  // Request adding a reasonable header, which ResourceFetch should accept:
+  EXPECT_TRUE(custom_options->ValidateAndAddResourceHeader(
+      "X-Resource-Header", kValue, &err));
+
+  // Separators should be accepted in the field value
+  EXPECT_TRUE(custom_options->ValidateAndAddResourceHeader(
+      "X-FooSeparator", "B; ar", &err));
+
+  // Values should be trimmed.
+  EXPECT_TRUE(custom_options->ValidateAndAddResourceHeader(
+      "  X-FooTrim  ", "  Bar   ", &err));
+
+  // Empty field value gets accepted.
+  EXPECT_TRUE(custom_options->ValidateAndAddResourceHeader(
+      "X-Foo-Spaced-Value", "aa bb", &err));
+
   RewriteDriver* custom_driver =
       server_context()->NewCustomRewriteDriver(
           custom_options,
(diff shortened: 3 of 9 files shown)


net/instaweb/rewriter/rewrite_options.cc
```diff
@@ -4432,6 +4442,124 @@ bool RewriteOptions::MergeOK() const {
...
+  // configured like that anyway.
+  TrimWhitespace(name, &gs_name);
+  TrimWhitespace(value, &gs_value);
+
+  if (ValidateConfiguredHttpHeader(gs_name, gs_value, error_message)) {
+    // We don't allow overwriting hop-by-hop headers like Connection.
+    StringPieceVector hbh = HttpAttributes::SortedHopByHopHeaders();
+    StringCompareInsensitive compare;
+    if (std::binary_search(&hbh[0], &hbh[0] + hbh.size(), gs_name, compare)) {
+        *error_message = StrCat("Rejecting hopy by hop header '", name, " '");
```
jmarantz: s/hopy/hop/

net/instaweb/rewriter/rewrite_options.cc
```diff
@@ -4432,6 +4442,124 @@ bool RewriteOptions::MergeOK() const {
...
+    DCHECK_EQ(kReserveSize, index);
+
+    if (std::binary_search(&tmp[0], &tmp[0] + tmp.size(), gs_name, compare)) {
+        *error_message = StrCat("Rejecting header '", name, " '");
+        return false;
+    }
+
+    // Arbitrary limit of adding 20 headers
+    if (resource_headers_.size() > 20) {
+      *error_message = "Too many AddResourceHeader directives";
```
jmarantz: Say in the error message what the maximum is.

net/instaweb/rewriter/public/rewrite_options.h
```diff
@@ -4220,6 +4247,8 @@ class RewriteOptions {
   int experiment_percent_;  // Total traffic going through experiments.
   std::vector<ExperimentSpec*> experiment_specs_;
 
+  // Headers to add resource responses.
+  std::vector<NameValue*> resource_headers_;
```
jmarantz, line 4259: consider wrapping this in a CopyOnWrite<NameValueVector> so that merges are free?

net/instaweb/rewriter/resource_fetch_test.cc
```diff
@@ -43,6 +43,7 @@ namespace {
 
 const char kCssContent[] = "* { display: none; }";
 const char kMinimizedCssContent[] = "*{display:none}";
+const GoogleString kValue("Value");
```
jmarantz: const char kValue[] = "Value";

We don't statically construct class instances, and in this case it's not needed anyway.

net/instaweb/rewriter/rewrite_options.cc
```diff
@@ -4432,6 +4442,124 @@ bool RewriteOptions::MergeOK() const {
...
+  }
+  if (name.size() > 1024) {
+    *error_message = "Field name too long";
+  }
+  if (value.size() > 1024) {
+    *error_message = "Field value too long";
+    return false;
+  }
+
+  for (size_t i = 0; i < name.size(); i++) {
```
jmarantz: ++i

net/instaweb/rewriter/public/rewrite_options.h
```diff
@@ -1043,6 +1044,32 @@ class RewriteOptions {
...
+
+  // If false was returned, no changes are made, and error_message will be
+  // assigned a failure description. If true was returned, the name/value
+  // passed validation and have been appended to resource_headers_
+  bool ValidateAndAddResourceHeader(const StringPiece& name,
+                                    const StringPiece& value,
+                                    GoogleString* error_message);
+
+  // Unconditionally appends the given name / value to resource_headers_. Use
+  // ValidateAndAddResourceHeader if you need validation.
```
jmarantz: If I have header a:b and then add header a:c, we'll wind up with both, right?  That's fine; but please doc it.

net/instaweb/rewriter/rewrite_options.cc
```diff
@@ -4432,6 +4442,124 @@ bool RewriteOptions::MergeOK() const {
...
+    tmp[index++] = StringPiece(HttpAttributes::kContentEncoding);
+    tmp[index++] = StringPiece(HttpAttributes::kContentLength);
+    tmp[index++] = StringPiece(HttpAttributes::kContentType);
+    tmp[index++] = StringPiece(HttpAttributes::kDate);
+    tmp[index++] = StringPiece(HttpAttributes::kEtag);
+    tmp[index++] = StringPiece(HttpAttributes::kExpires);
+    tmp[index++] = StringPiece(HttpAttributes::kLastModified);
+    tmp[index++] = StringPiece(HttpAttributes::kLink);
+    tmp[index++] = StringPiece(HttpAttributes::kServer);
+    tmp[index++] = StringPiece(HttpAttributes::kVary);
```
jmarantz: Can you initialize this StringPieceVector as an anon-namespace constant in RewriteOptions::Initialize?

net/instaweb/rewriter/public/rewrite_options.h
oschaaf, line 4259: re: "consider wrapping this in a CopyOnWrite so that merges are free?"

will do, should I also do so for `custom_fetch_headers_` one line below while I'm at it?

oschaaf, line 4259: @jmarantz 
I'm looking into wrapping RewriteOptions::resource_headers_ in a CopyOnWrite as suggested. The thing is that currently RewriteOptions::resource_headers_ is just a std::vector<NameValue*>, which doesn't have the required Merge() method for CopyOnWrite to work with.

I think it's worth factoring out a class which has the required Merge() method + validations implemented, especially as we can also reuse that for CustomFetchHeaders.

Questions: 
- Does that sound good? 
- If so, do you want me to do so now, or first target merging without wrapping CopyOnWrite, and subsequently refactor / wrap CopyOnWrap in a second pull? A second pull might also be a good place to address a few small things for CustomFetchHeaders (move its system tests to shared, share validations and CopyOnWrap usage with AddResourceHeaders through the new class)?

jmarantz, line 4259: Yes, doing this in a follow-up pull-request makes sense.  Please leave a TODO indicating your intention.  Thanks!!

oschaaf:
@jmarantz I addressed your comments, PTAL?

net/instaweb/rewriter/rewrite_options.cc
```diff
@@ -995,6 +996,9 @@ std::vector<DeprecatedOptionMap> kDeprecatedOptionNameList(
     kDeprecatedOptionNameData + arraysize(kDeprecatedOptionNameData)
 );
 
+int kNonAddableResourceHeadersSize = 12;
```
jmarantz: You don't really need this or the DCHECK which is the only place it's referenced.

net/instaweb/rewriter/rewrite_options.cc
```diff
@@ -995,6 +996,9 @@ std::vector<DeprecatedOptionMap> kDeprecatedOptionNameList(
...
+StringPieceVector kNonAddableResourceHeaders(kNonAddableResourceHeadersSize);
```
jmarantz: Sorry to be a pain; this needs to be a pointer which you allocate in InitNonAddableResourceHeaders and delete in Terminate, so you don't run a static initializer.

jmarantz:
Sorry I put in a couple of notes but I am actually taking the day off today
and if I'm not careful I'll wind up working most of it:)

I'll finish the re-review tomorrow.  Cool?

oschaaf:
@jmarantz Sounds good. Meantime, I'll address your new comments. Have a nice day off!

oschaaf:
@jmarantz updated & force pushed.

net/instaweb/rewriter/rewrite_options.cc
oschaaf: @jmarantz Out of curiosity, is kDeprecatedOptionNameList (declared one line above) valid according to the style guide? That line was why I thought that what I had initially was OK.

jmarantz: kDeprecatedOptionNameList is definitely not valid.  Sorry about the
confusion; that slipped through the review process.

jeffkaufman:
@jmarantz any more comments?  LGTY?

jmarantz:
sorry all; fell off my radar.  I'm looking now.

net/instaweb/rewriter/public/rewrite_options.h
```diff
@@ -1043,6 +1044,34 @@ class RewriteOptions {
     set_option(level, &level_);
   }
 
+  // returns true iff given name and value are valid to set as a http header.
```
jmarantz: Returns

net/instaweb/rewriter/resource_fetch.cc
```diff
@@ -171,6 +171,10 @@ void ResourceFetch::HandleHeadersComplete() {
   DCHECK(!response_headers()->Lookup(HttpAttributes::kSetCookie2, &v));
   response_headers()->RemoveAll(HttpAttributes::kSetCookie);
   response_headers()->RemoveAll(HttpAttributes::kSetCookie2);
+  for (int i = 0; i < driver_->options()->num_resource_headers(); i++) {
```
jmarantz: ++i

net/instaweb/rewriter/rewrite_options.cc
```diff
@@ -2641,7 +2647,29 @@ void RewriteOptions::InitOptionNameToPropertyArray() {
   }
 }
 
+void RewriteOptions::InitFixedResourceHeaders() {
+  int index = 0;
+  StringPieceVector* tmp = new StringPieceVector(12);
```
jmarantz: i probably wouldn't bother with estimating the size here; just say 'new StringPieceVector;', and use push_back below.  This will only be run once on process startup.

net/instaweb/rewriter/rewrite_options.cc
```diff
@@ -4432,6 +4468,108 @@ bool RewriteOptions::MergeOK() const {
...
+    *error_message = "Field value too long";
+    return false;
+  }
+
+  for (size_t i = 0; i < name.size(); ++i) {
+    if (!IsNonControlAscii(name[i])) {
+      *error_message = "Invalid character in field name";
+      return false;
+    }
+    if (separators.find(name[i]) != GoogleString::npos) {
```
jmarantz: This looks slow. How about static-initing an ascii-ordered char[] kSeparators and using binary-search on that, or use a 256-element bool array that's initialized where you set up the string-vector array above.

net/instaweb/rewriter/rewrite_options.cc
```diff
@@ -4432,6 +4468,108 @@ bool RewriteOptions::MergeOK() const {
...
+      *error_message = "Separator found in field name";
+      return false;
+    }
+  }
+
+  for (size_t i = 0; i < value.size(); ++i) {
+    // Formally the value may have any OCTET except CNTL, but including LWS.
+    // But in practice I think this is fair enough for our use case:
+    if (!IsNonControlAscii(value[i])) {
+      *error_message = "Invalid character in field value";
```
jmarantz: append the invalid separator, so whoever violated it knows which one it was.

oschaaf:
@jmarantz All comments addressed, updated pull.
The field separators are now pulled from a 256-element bool array.

net/instaweb/rewriter/rewrite_options.cc
```diff
@@ -995,6 +996,11 @@ std::vector<DeprecatedOptionMap> kDeprecatedOptionNameList(
     kDeprecatedOptionNameData + arraysize(kDeprecatedOptionNameData)
 );
 
+// Will be initialized to a sorted list of headers not allowed in
+// AddResourceHeader.
+const StringPieceVector* fixed_resource_headers_ = NULL;
+bool http_header_separators_[256] = {false};
```
jmarantz: Remove the trailing underscore -- those are for member vars and do not connote private or file-local.

I am not aware of any way to distinguish namespaced constants from locals via naming convention.

pagespeed/system/system_test.sh
```diff
@@ -2249,3 +2249,10 @@ if [ "$CACHE_FLUSH_TEST" = "on" ]; then
   # check [ $bytes -gt 100000 ]
   check [ $bytes -lt 100000 ]
 fi
+
+start_test AddResourceHeaders works for pagespeed resources.
+URL="$TEST_ROOT/compressed/hello_js.custom_ext.pagespeed.ce.HdziXmtLIV.txt"
+HTML_HEADERS=$($WGET_DUMP $URL)
+# TODO(oschaaf): a helper to perform this type of matching.
+MATCHES=$(echo "$HTML_HEADERS" | grep -c "^X-Foo: Bar") || true
```
jmarantz: how about:
  check_from "$HTML_HEADERS" grep -q "^X-Foo: Bar"
See also extract_headers, which will give you just the headers, but it requires a file.

The advantage of check_from is that if it fails, it prints what the contents of the variable in the failure message.  This is important for flaky tests, where if you simply try it again it will not necessarily fail, or fail the same way.

net/instaweb/rewriter/rewrite_options.cc
oschaaf: Done & pushed. Just to make clear, the line above is ok? (`const StringPieceVector* fixed_resource_headers_`)?

pagespeed/system/system_test.sh
oschaaf: I'll use check_from. Does check_from also ensure there is single instance of the header? I think that is the common case, and in nps we currently have, for example, multiple vary headers being emitted.

oschaaf:
@jmarantz comments addressed again

net/instaweb/rewriter/rewrite_options.cc
jmarantz: Sorry -- the comment was meant to apply to both lines :)

oschaaf: @jmarantz re:"Sorry -- the comment was meant to apply to both lines :)"
Fixed

jmarantz:
LGTM

RE check_from and counts -- I know we have counts elsewhere basically that function captures $1, shifts, and echos the captured text into $@.

So we can put whatever function we want e.g.
  check_from "$OUT" grep_check_count pattern count
if we define
  function grep_check_count() {
    count=$(grep "$1")
    if [ $count -ne "$2" ]; then
...

something like that.  My bash is not quite fluent to do that off the top of my head :)

jmarantz:
See also automatic/system_test_helpers.sh where check_from is defined.  More helpers can be added there.

PR #1223, opened 2015-12-17
build with gcc 4.8 on travis

crowell:
wait for this to pass in travis before merging I think this fixes the issue, but there may be a few more reasons why the build is failing currently.


crowell:
gcc build passed, clang still is broken with a linker error, working on that and merging this in for now.

PR #1250, opened 2016-01-21
Move InstawebHandler initialization after validity checks.

jeffkaufman:
Creating an instaweb handler will run MakeRequestUrl, which assumes that
request->unparsed_uri is non-null.  So move the creation to after where
we check that it's non-null.  To be safe, move it all the way down to
where it's first needed, in case some other validity checks end up being
relevant.

Fixes https://github.com/pagespeed/mod_pagespeed/issues/1248

### DIFF
--- a/pagespeed/apache/mod_instaweb.cc
+++ b/pagespeed/apache/mod_instaweb.cc
@@ -378,9 +378,6 @@ InstawebContext* build_context_for_request(request_rec* request) {
     return NULL;
   }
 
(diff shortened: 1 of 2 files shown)


oschaaf:
Makes sense, LGTM

jeffkaufman:
Merged as https://github.com/pagespeed/mod_pagespeed/commit/befa49422be64d0ab69aa5adf79005c9ec744f73

PR #1255, opened 2016-01-27
Use fetch_until for AddResourceHeaders check

jmarantz:
The fetch in that test depends on Serf, which can timeout during valgrind tests.


oschaaf:
LGTM - sorry you ran into that.

PR #1260, opened 2016-02-09
Add centralized process

jeffkaufman:
This change adds two processes: a controller, and a babysitter.  The
controller process is intended to host cheesy's controller work.  The
babysitter process just waits around in case the controller dies and
restarts it if it does.

In order to exit when the host process exits, the controller watches a pipe.
If reading from that pipe gets an EOF, then it knows the master quit and it
can quit itself.  If the master wants to quit the controller to load a new
one with an updated configuration, it can write a byte to the pipe, which
the controller will see and exit.

### DIFF
--- a/pagespeed/apache/mock_apache.cc
+++ b/pagespeed/apache/mock_apache.cc
@@ -1,5 +1,9 @@
 #include "pagespeed/apache/mock_apache.h"
 
+#include <cstdlib>
+#include <vector>
+
+#include "base/logging.h"
 #include "pagespeed/kernel/base/string_util.h"
 #include "pagespeed/kernel/http/http_names.h"
 
@@ -159,6 +163,13 @@ ap_filter_rec_t* ap_register_input_filter(const char*, ap_in_filter_func,
   log_fatal("ap_register_input_filter");
   return NULL;
 }
+
+const char* ap_server_root_relative(apr_pool_t*, const char*) {
+  log_fatal("ap_server_root_relative");
+  return NULL;
+}
+
+const char* ap_pid_fname = NULL;
 
 #define IMPLEMENT_AS_LOG_FATAL(AP_X) \
   void AP_X() { log_fatal(#AP_X); }
--- a/pagespeed/apache/mock_apache.cc
+++ b/pagespeed/apache/mock_apache.cc
@@ -164,13 +164,6 @@ ap_filter_rec_t* ap_register_input_filter(const char*, ap_in_filter_func,
   return NULL;
 }
 
-const char* ap_server_root_relative(apr_pool_t*, const char*) {
-  log_fatal("ap_server_root_relative");
-  return NULL;
-}
-
-const char* ap_pid_fname = NULL;
-
 #define IMPLEMENT_AS_LOG_FATAL(AP_X) \
   void AP_X() { log_fatal(#AP_X); }
 
--- a/pagespeed/system/controller_manager.cc
+++ b/pagespeed/system/controller_manager.cc
@@ -1,0 +1,228 @@
+// Copyright 2016 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Author: jefftk@google.com (Jeff Kaufman)
+
+#include "pagespeed/system/controller_manager.h"
+
+#include <signal.h>
+#include <sys/wait.h>
+#include <unistd.h>
+#include <cerrno>
+#include <cstddef>
+#include <cstdlib>
+
+#include "base/logging.h"
+#include "pagespeed/kernel/base/basictypes.h"
+#include "pagespeed/kernel/base/string_util.h"
+#include "pagespeed/system/system_rewrite_options.h"
+
+namespace net_instaweb {
+
+
+ControllerManager::ControllerManager(
+    SystemRewriteDriverFactory* factory,
+    ThreadSystem* thread_system,
+    MessageHandler* handler) : factory_(factory),
+                               thread_system_(thread_system),
+                               handler_(handler) { }
+
+void ControllerManager::SetUpSignalHandlers() {
+  // We need to clear inherited signal handlers. There's no portable way to get
+  // a list of all possible signals, and they're not even guaranteed to be in
+  // order.  But NSIG is usually defined these days, and if it is then we just
+  // want ascending numbers up to to NSIG.
+  for (int i = 0; i < NSIG; i++) {
+    signal(i, SIG_DFL);
+  }
+}
+
+ControllerManager::ProcessDeathWatcherThread::ProcessDeathWatcherThread(
+    ThreadSystem* thread_system,
+    int controller_read_fd,
+    MessageHandler* handler) : Thread(thread_system,
+                                      "process death watcher",
+                                      ThreadSystem::kDetached),
+                              handler_(handler),
+                              controller_read_fd_(controller_read_fd) {
+}
+
+void ControllerManager::ProcessDeathWatcherThread::Run() {
+  handler_->Message(kInfo, "Watching the root process to exit if it does.");
+
+  ssize_t status;
+  char buf[1];
+  do {
+    status = read(controller_read_fd_, buf, 1);
+  } while (status == -1 && (errno == EINTR ||
+                            errno == EAGAIN));
+  if (status == -1) {
+    handler_->Message(
+        kWarning, "Controller got error %d reading from pipe, shutting down",
+        static_cast<int>(status));
+  } else if (status == 0 /* EOF */) {
+    handler_->Message(kInfo, "Root process exited; controller shutting down.");
+  } else if (status == 1 /* read a byte */) {
+    handler_->Message(
+        kInfo, "Root process is starting a new controller; shutting down.");
+  } else {
+    LOG(FATAL) << "Status of " << status << " doesn't make sense";
+  }
+
+  // TODO(jefftk): Once the controller is doing real work and there's real
+  // cleanup work to do, we can initiate that from here.
+
+  exit(EXIT_SUCCESS);
+
+  // Because the controller is exiting via exit(), the babysitter process will
+  // see that and quit itself instead of restarting it.
+}
+
+void ControllerManager::RunController() {
+  handler_->Message(kInfo, "Controller running with PID %d", getpid());
+
+  // Would set up gRPC server here and pass control to its event loop.  Instead
+  // just hang out sleeping until the ProcessDeathWatcherThread decides it's
+  // time to exit.
+  while (true) {
+    sleep(60);  // 1m
+  }
+}
+
+void ControllerManager::Daemonize() {
+  // Make a new session (process group).
+  if (setsid() < 0) {
+    handler_->Message(kWarning, "Daemonize: Failed to setsid().");
+  }
+
+  // We need to fork again to make sure there is no session group leader.
+  pid_t pid = fork();
+  CHECK(pid != -1) << "Couldn't fork to daemonize.";
+  if (pid != 0) {
+    exit(EXIT_SUCCESS);
+  }
+
+  // If we keep the current directory we might keep them from being able to
+  // unmount their filesystem.
+  if (chdir("/") < 0) {
+    handler_->Message(kWarning, "Daemonize: Failed to chdir(/).");
+  }
+
+  // If we disconnect file descriptors then logging will break, so don't.
+}
+
+void ControllerManager::ForkOffControllerProcess() {
+  handler_->Message(kInfo, "Forking controller process off of %d", getpid());
+
+  // Whenever we fork off a controller we save the fd for a pipe to it.  Then if
+  // we fork off another controller we can write a byte to the pipe to tell the
+  // old controller to clean up and exit.
+  static int controller_write_fd = -1;
+  if (controller_write_fd != -1) {
+    // We already forked off a controller earlier.  Tell it to quit by writing a
+    // byte.  If there's no one still with the pipe open we'll get SIGPIPE and
+    // die horribly, but the babysitter-restarting-controller structure makes
+    // that very unlikely.
+    handler_->Message(
+        kInfo, "Writing a byte to a pipe to tell the old controller to exit.");
+    ssize_t status;
+    do {
+      status = write(controller_write_fd, "Q", 1);
+    } while (status == -1 && (errno == EAGAIN ||
+                              errno == EINTR));
+    if (status == -1) {
+      handler_->Message(kWarning, "killing old controller failed: %s",
+                        strerror(errno));
+    }
+  }
+
+  int file_descriptors[2];
+  int pipe_status = pipe(file_descriptors);
+  CHECK(pipe_status != -1) << "Couldn't create a root-controller pipe.";
+
+  pid_t pid = fork();
+  CHECK(pid != -1) << "Couldn't fork a controller babysitter process";
+
+  if (pid != 0) {
+    // Parent process.
+
+    // Close the reading end of the pipe.  We'll never write to it, but when we
+    // (and all our children) die there will be no more processes that could
+    // potentially write to it, and so the people who do have it open for
+    // reading can see that death.
+    close(file_descriptors[0]);
+
+    // Save the writing end of the pipe.
+    controller_write_fd = file_descriptors[1];
+
+    return;
+  }
+
+  // Now we're in the child process.  Set this up as a babysitter process,
+  // that forks off a controller and restarts it if it dies.
+
+  Daemonize();
+  SetUpSignalHandlers();
+
+  factory_->set_is_root_process(false);
+  factory_->PrepareForkedProcess("babysitter");
+
+  // Close the writing end of the pipe.  If we read a byte from the pipe it
+  // means we should quit because a new controller is starting up.  If we get
+  // EOF from the pipe it means we should quit because the master process shut
+  // down.
+  close(file_descriptors[1]);
+  int controller_read_fd = file_descriptors[0];
+
+  handler_->Message(kInfo, "Babysitter running with PID %d", getpid());
+
+  while (true) {
+    pid = fork();
+    CHECK(pid != -1) << "Couldn't fork a controller process";
+
+    if (pid == 0) {
+      factory_->PrepareForkedProcess("controller");
+      factory_->PrepareControllerProcess();
+
+      // Start a thread to watch to see if the root or babysitter process
+      // dies, and quit if it does.
+      process_death_watcher_thread_.reset(new ProcessDeathWatcherThread(
+          thread_system_, controller_read_fd, handler_));
+      CHECK(process_death_watcher_thread_->Start());
+
+      RunController();
+    } else {
+      // Wait for controller process to die, then continue with the loop by
+      // restarting it.
+      int status;
+      pid_t child_pid;
+      do {
+        child_pid = waitpid(pid, &status, 0);
+      } while (child_pid == -1 && errno == EINTR);
+      CHECK(child_pid != -1) << "Call to waitpid failed with status "
+                             << child_pid;
+      if (WIFEXITED(status)) {
+        handler_->Message(kInfo,
+            "Controller process %d exited normally, not restarting it. "
+            "Shutting down babysitter.", child_pid);
+        exit(EXIT_SUCCESS);
+      }
+      handler_->Message(
+          kWarning, "Controller process %d exited with status code %d",
+          child_pid, status);
+    }
+  }
+}
+
+}  // namespace net_instaweb
--- a/pagespeed/system/controller_manager.cc
+++ b/pagespeed/system/controller_manager.cc
@@ -170,7 +170,6 @@ void ControllerManager::ForkOffControllerProcess(
     signal(i, SIG_DFL);
   }
 
-  factory->set_is_root_process(false);
   factory->PrepareForkedProcess("babysitter");
 
   // Close the writing end of the pipe.  If we read a byte from the pipe it
@@ -199,6 +198,7 @@ void ControllerManager::ForkOffControllerProcess(
       CHECK(process_death_watcher_thread->Start());
 
       RunController(factory, handler);
+      LOG(FATAL) << "Controller should run until exit().";
     } else {
       // Wait for controller process to die, then continue with the loop by
       // restarting it.
--- a/pagespeed/system/controller_manager.cc
+++ b/pagespeed/system/controller_manager.cc
@@ -20,13 +20,11 @@
 #include <sys/wait.h>
 #include <unistd.h>
 #include <cerrno>
-#include <cstddef>
 #include <cstdlib>
 
 #include "base/logging.h"
-#include "pagespeed/kernel/base/basictypes.h"
-#include "pagespeed/kernel/base/string_util.h"
-#include "pagespeed/system/system_rewrite_options.h"
+#include "third_party/pagespeed/kernel/base/scoped_ptr.h"
+#include "third_party/pagespeed/kernel/base/string.h"
 
 namespace net_instaweb {
 
@@ -37,8 +35,8 @@ ControllerManager::ProcessDeathWatcherThread::ProcessDeathWatcherThread(
     MessageHandler* handler) : Thread(thread_system,
                                       "process death watcher",
                                       ThreadSystem::kDetached),
-                              handler_(handler),
-                              controller_read_fd_(controller_read_fd) {
+                               handler_(handler),
+                               controller_read_fd_(controller_read_fd) {
 }
 
 void ControllerManager::ProcessDeathWatcherThread::Run() {
--- a/pagespeed/system/controller_manager.cc
+++ b/pagespeed/system/controller_manager.cc
@@ -31,23 +31,6 @@
 namespace net_instaweb {
 
 
-ControllerManager::ControllerManager(
-    SystemRewriteDriverFactory* factory,
-    ThreadSystem* thread_system,
-    MessageHandler* handler) : factory_(factory),
-                               thread_system_(thread_system),
-                               handler_(handler) { }
-
-void ControllerManager::SetUpSignalHandlers() {
-  // We need to clear inherited signal handlers. There's no portable way to get
-  // a list of all possible signals, and they're not even guaranteed to be in
-  // order.  But NSIG is usually defined these days, and if it is then we just
-  // want ascending numbers up to to NSIG.
-  for (int i = 0; i < NSIG; i++) {
-    signal(i, SIG_DFL);
-  }
-}
-
 ControllerManager::ProcessDeathWatcherThread::ProcessDeathWatcherThread(
     ThreadSystem* thread_system,
     int controller_read_fd,
@@ -89,8 +72,9 @@ void ControllerManager::ProcessDeathWatcherThread::Run() {
   // see that and quit itself instead of restarting it.
 }
 
-void ControllerManager::RunController() {
-  handler_->Message(kInfo, "Controller running with PID %d", getpid());
+void ControllerManager::RunController(SystemRewriteDriverFactory* factory,
+                                      MessageHandler* handler) {
+  handler->Message(kInfo, "Controller running with PID %d", getpid());
 
   // Would set up gRPC server here and pass control to its event loop.  Instead
   // just hang out sleeping until the ProcessDeathWatcherThread decides it's
@@ -100,10 +84,10 @@ void ControllerManager::RunController() {
   }
 }
 
-void ControllerManager::Daemonize() {
+void ControllerManager::Daemonize(MessageHandler* handler) {
   // Make a new session (process group).
   if (setsid() < 0) {
-    handler_->Message(kWarning, "Daemonize: Failed to setsid().");
+    handler->Message(kWarning, "Daemonize: Failed to setsid().");
   }
 
   // We need to fork again to make sure there is no session group leader.
@@ -116,14 +100,18 @@ void ControllerManager::Daemonize() {
   // If we keep the current directory we might keep them from being able to
   // unmount their filesystem.
   if (chdir("/") < 0) {
-    handler_->Message(kWarning, "Daemonize: Failed to chdir(/).");
+    handler->Message(kWarning, "Daemonize: Failed to chdir(/).");
   }
 
   // If we disconnect file descriptors then logging will break, so don't.
 }
 
-void ControllerManager::ForkOffControllerProcess() {
-  handler_->Message(kInfo, "Forking controller process off of %d", getpid());
+void ControllerManager::ForkOffControllerProcess(
+    SystemRewriteDriverFactory* factory,
+    ThreadSystem* thread_system,
+    MessageHandler* handler) {
+
+  handler->Message(kInfo, "Forking controller process off of %d", getpid());
 
   // Whenever we fork off a controller we save the fd for a pipe to it.  Then if
   // we fork off another controller we can write a byte to the pipe to tell the
@@ -132,9 +120,9 @@ void ControllerManager::ForkOffControllerProcess() {
   if (controller_write_fd != -1) {
     // We already forked off a controller earlier.  Tell it to quit by writing a
     // byte.  If there's no one still with the pipe open we'll get SIGPIPE and
-    // die horribly, but the babysitter-restarting-controller structure makes
-    // that very unlikely.
-    handler_->Message(
+    // die horribly, but as long as the babysitter hasn't died that won't
+    // happen.
+    handler->Message(
         kInfo, "Writing a byte to a pipe to tell the old controller to exit.");
     ssize_t status;
     do {
@@ -142,8 +130,8 @@ void ControllerManager::ForkOffControllerProcess() {
     } while (status == -1 && (errno == EAGAIN ||
                               errno == EINTR));
     if (status == -1) {
-      handler_->Message(kWarning, "killing old controller failed: %s",
-                        strerror(errno));
+      handler->Message(kWarning, "killing old controller failed: %s",
+                       strerror(errno));
     }
   }
 
@@ -172,11 +160,18 @@ void ControllerManager::ForkOffControllerProcess() {
   // Now we're in the child process.  Set this up as a babysitter process,
   // that forks off a controller and restarts it if it dies.
 
-  Daemonize();
-  SetUpSignalHandlers();
+  Daemonize(handler);
 
-  factory_->set_is_root_process(false);
-  factory_->PrepareForkedProcess("babysitter");
+  // We need to clear inherited signal handlers. There's no portable way to get
+  // a list of all possible signals, and they're not even guaranteed to be in
+  // order.  But NSIG is usually defined these days, and if it is then we just
+  // want ascending numbers up to to NSIG.
+  for (int i = 0; i < NSIG; i++) {
+    signal(i, SIG_DFL);
+  }
+
+  factory->set_is_root_process(false);
+  factory->PrepareForkedProcess("babysitter");
 
   // Close the writing end of the pipe.  If we read a byte from the pipe it
   // means we should quit because a new controller is starting up.  If we get
@@ -185,23 +180,25 @@ void ControllerManager::ForkOffControllerProcess() {
   close(file_descriptors[1]);
   int controller_read_fd = file_descriptors[0];
 
-  handler_->Message(kInfo, "Babysitter running with PID %d", getpid());
+  handler->Message(kInfo, "Babysitter running with PID %d", getpid());
 
   while (true) {
     pid = fork();
     CHECK(pid != -1) << "Couldn't fork a controller process";
 
     if (pid == 0) {
-      factory_->PrepareForkedProcess("controller");
-      factory_->PrepareControllerProcess();
+      factory->PrepareForkedProcess("controller");
+      factory->PrepareControllerProcess();
 
       // Start a thread to watch to see if the root or babysitter process
       // dies, and quit if it does.
-      process_death_watcher_thread_.reset(new ProcessDeathWatcherThread(
-          thread_system_, controller_read_fd, handler_));
-      CHECK(process_death_watcher_thread_->Start());
+      scoped_ptr<ProcessDeathWatcherThread> process_death_watcher_thread(
+          new ProcessDeathWatcherThread(thread_system,
+                                        controller_read_fd,
+                                        handler));
+      CHECK(process_death_watcher_thread->Start());
 
-      RunController();
+      RunController(factory, handler);
     } else {
       // Wait for controller process to die, then continue with the loop by
       // restarting it.
@@ -213,12 +210,12 @@ void ControllerManager::ForkOffControllerProcess() {
       CHECK(child_pid != -1) << "Call to waitpid failed with status "
                              << child_pid;
       if (WIFEXITED(status)) {
-        handler_->Message(kInfo,
+        handler->Message(kInfo,
             "Controller process %d exited normally, not restarting it. "
             "Shutting down babysitter.", child_pid);
         exit(EXIT_SUCCESS);
       }
-      handler_->Message(
+      handler->Message(
           kWarning, "Controller process %d exited with status code %d",
           child_pid, status);
     }
--- a/pagespeed/system/controller_manager.cc
+++ b/pagespeed/system/controller_manager.cc
@@ -106,7 +106,7 @@ void ControllerManager::Daemonize(MessageHandler* handler) {
   // If we disconnect file descriptors then logging will break, so don't.
 }
 
-void ControllerManager::ForkOffControllerProcess(
+void ControllerManager::ForkControllerProcess(
     SystemRewriteDriverFactory* factory,
     ThreadSystem* thread_system,
     MessageHandler* handler) {
--- a/pagespeed/system/controller_manager.cc
+++ b/pagespeed/system/controller_manager.cc
@@ -53,7 +53,7 @@ void ControllerManager::ProcessDeathWatcherThread::Run() {
   if (status == -1) {
     handler_->Message(
         kWarning, "Controller got error %d reading from pipe, shutting down",
-        static_cast<int>(status));
+        errno);
   } else if (status == 0 /* EOF */) {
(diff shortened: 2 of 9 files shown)


jeffkaufman:
nginx side of the change: https://github.com/pagespeed/ngx_pagespeed/pull/1113

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,335 @@
...
+    MessageHandler* handler) : factory_(factory),
+                               thread_system_(thread_system),
+                               file_system_(file_system),
+                               timer_(timer),
+                               handler_(handler),
+                               babysitter_pid_(-1) { }
+
+void ControllerManager::SetUpSignalHandlers() {
+  // We need to clear inherited signal handlers. There's no portable way to get
+  // a list of all possible signals, but these are the rest of the ones that we
```
crowell: we can get this data from `signal.h`, and use the macro for `NSIG`.

Since the signal numbers are allocated consecutively, `NSIG` is also one greater than the largest defined signal number.

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,335 @@
...
+  // We need to fork again to make sure there is no session group leader.
+  pid_t pid = fork();
+  CHECK(pid != -1) << "Couldn't fork to daemonize.";
+  if (pid != 0) {
+    exit(EXIT_SUCCESS);
+  }
+
+  // If we keep the current directory we might keep them from being able to
+  // unmount their filesystem.
+  chdir("/");
```
crowell: check for error on `chdir` (build warning here)

```
pagespeed/system/controller_manager.cc: In member function â€˜void net_instaweb::ControllerManager::Daemonize()â€™:
pagespeed/system/controller_manager.cc:268:13: error: ignoring return value of â€˜int chdir(const char*)â€™, declared with attribute warn_unused_result [-Werror=unused-result]
   chdir("/");
```

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,335 @@
...
+        kError, "Controller PID Filename should not be set to empty");
+    return "";
+  } else if (controller_pid_filename[0] == '/') {
+    return controller_pid_filename;
+  } else {
+    // Relative paths are interpreted relative to the directory in which the
+    // webserver would normally store the pid file for its master process.
+
+    GoogleString master_pid_filename = factory_->MasterPidFilename();
+    size_t final_slash = master_pid_filename.find_last_of('/');
```
crowell: `dirname(3)`

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,335 @@
...
+    if (errno == ESRCH) {
+      // This is fine.  This is what usually happens if the controller exits
+      // abruptly and leaves behind a stale pid in its file.
+    } else if (errno == EPERM) {
+      handler_->Message(
+          kWarning,
+          "Permissions error trying to kill controller process at PID %d.",
+          controller_pid);
+    } else {
+      LOG(DFATAL) << "Got unexpected failure " << errno << " from kill.";
```
crowell: `strerror` here might be helpful?

pagespeed/system/controller_manager.cc
jeffkaufman: Done.  NSIG isn't posix, but it seems to be widely supported.

pagespeed/system/controller_manager.cc
jeffkaufman: Added log messages for both chdir and setsid failure.

pagespeed/system/controller_manager.cc
jeffkaufman: `dirname` is evil.  It can modify it's input, segfault on static strings, and isn't thread safe.

pagespeed/system/controller_manager.cc
jeffkaufman: We should only ever get ESRCH or EPERM; kill isn't documented as returning any others.  So I'm happy without strerror.

net/instaweb/rewriter/public/rewrite_options.h
```diff
@@ -236,6 +236,7 @@ class RewriteOptions {
   static const char kCombineAcrossPaths[];
   static const char kContentExperimentID[];
   static const char kContentExperimentVariantID[];
+  static const char kControllerPidFilename[];
```
morlovich: Hmm, why have this here and not system_rewrite_options?

pagespeed/system/system_rewrite_options.cc
```diff
@@ -245,6 +245,12 @@ void SystemRewriteOptions::AddProperties() {
                     "URLs rather than from compiled-in versions via static "
                     "handler.", true);
 
+  AddSystemProperty("pagespeed_controller.pid",
+                    &SystemRewriteOptions::controller_pid_filename_,
+                    "cpf",
+                    RewriteOptions::kControllerPidFilename, kProcessScope,
+                    "File to store our pid in.", false);
```
morlovich: What's our here?

pagespeed/apache/apache_rewrite_driver_factory.cc
```diff
@@ -227,4 +236,18 @@ ApacheConfig* ApacheRewriteDriverFactory::NewRewriteOptionsForQuery() {
   return new ApacheConfig("query", thread_system());
 }
 
+GoogleString ApacheRewriteDriverFactory::MasterPidFilename() {
```
morlovich: It seems like you use 'master' here for the same thing a lot of the code uses 'root' for, maybe worth considering using the same terminology?

pagespeed/apache/apache_rewrite_driver_factory.cc
```diff
@@ -227,4 +236,18 @@ ApacheConfig* ApacheRewriteDriverFactory::NewRewriteOptionsForQuery() {
...
+  if (ap_pid_fname == NULL) {
+    LOG(DFATAL) << "ap_pid_fname should not be null";
+    return "";
+  }
+  const char* fname = ap_server_root_relative(pool_, ap_pid_fname);
```
morlovich: Hmm, this is potentially slightly leaky if this gets called a lot, which it probably isn't.

pagespeed/apache/mock_apache.cc
```diff
@@ -6,6 +6,10 @@
 #include "httpd.h"        // NOLINT
 #include "util_filter.h"  // NOLINT
 
+#include <cstdlib>
+#include <vector>
+#include "base/logging.h"
```
morlovich: This should probably go towards line 3 or so. Or is there some #define awkwardness that's getting in the way?

pagespeed/system/controller_manager.h
```diff
@@ -0,0 +1,141 @@
...
+#include "pagespeed/kernel/base/string.h"
+#include "pagespeed/kernel/base/thread.h"
+#include "pagespeed/kernel/base/thread_system.h"
+#include "pagespeed/kernel/base/timer.h"
+#include "pagespeed/system/system_rewrite_driver_factory.h"
+
+
+namespace net_instaweb {
+
+// A server context with features specific to a psol port on a unix system.
```
morlovich: Odd comment.

pagespeed/system/controller_manager.h
```diff
@@ -0,0 +1,141 @@
...
+                    Timer* timer,
+                    MessageHandler* handler);
+  virtual ~ControllerManager() {}
+
+  // Called on system startup, before forking off any workers.  Starts up a
+  // babysitter process that starts a controller process and restarts the
+  // controller if it dies.  Also called (again) on configuration reloading.
+  void ForkOffControllerProcess();
+
+  class ProcessDeathWatcherThread : public ThreadSystem::Thread {
```
morlovich, line 67: Any need for this to be public?

pagespeed/system/controller_manager.h
```diff
@@ -0,0 +1,141 @@
...
+    pid_t babysitter_pid_;
+
+    const char* controller_pid_filename_;
+
+    DISALLOW_COPY_AND_ASSIGN(ProcessDeathWatcherThread);
+  };
+
+ private:
+  // Controller will be hooked up here.  This method is called in a single
+  // centralized process, and if that process dies it will be started again.
```
morlovich: is "a single centralized process" the babysitter, or something else?

pagespeed/system/controller_manager.h
```diff
@@ -0,0 +1,141 @@
...
+  pid_t ReadPidFromFile(const GoogleString& filename);
+
+  // Look up the PID of the controller process.  If the controller process is
+  // not running, returns -1 immediately.
+  pid_t ControllerPid();
+
+  // TODO(jefftk): This is confusing verging on misleading.  This actually
+  // writes the babysitter pid, and the way we kill the controller is to kill
+  // the babysitter.  So the "controller pid file" actually contains the pid of
+  // the babysitter process.
```
morlovich: Not sure why you would want to punt on naming stuff like this from this CL...

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,329 @@
...
+  return kill(pid, 0) == 0;
+}
+
+void ControllerManager::ProcessDeathWatcherThread::Run() {
+  handler_->Message(
+      kInfo, "Watching master (%d) and babysitter (%d) to exit if they do.",
+      master_pid_, babysitter_pid_);
+  while (IsProcessRunning(master_pid_) &&
+         IsProcessRunning(babysitter_pid_)) {
+    timer_->SleepUs(30 * 1000 /* 0.03s */);
```
morlovich: That's a lot of wakeups. Any way of using a wait\* method?

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,329 @@
...
+  // cleanup work to do, we can initiate that from here.
+
+  exit(EXIT_SUCCESS);
+
+  // Because the controller is exiting via exit(), the babysitter process will
+  // see that and quit itself instead of restarting it.
+}
+
+pid_t ControllerManager::ReadPidFromFile(
+    const GoogleString& filename) {
```
morlovich: Unwrap?

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,329 @@
...
+  if (file_system_->Exists(
+          filename.c_str(), handler_).is_true()) {
+    GoogleString pid_buffer;
```
morlovich: Why the explicit exists? (Partly asking since the file can disappear between exists and read... though of course it also can after the read)

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,329 @@
...
+    }
+
+    // Sleep instead of spinning.
+    timer_->SleepUs(10 * 1000 /* 0.01s */);
+  }
+}
+
+GoogleString ControllerManager::ControllerPidFilename() {
+  SystemRewriteOptions* options = reinterpret_cast<SystemRewriteOptions*>(
+      factory_->default_options());  // Cast is safe, since we're in system.
```
morlovich: It still shouldn't be a reinterpret_cast, but a static_cast (or dynamic_cast, given your check below)

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,329 @@
...
+  } else {
+    // Relative paths are interpreted relative to the directory in which the
+    // webserver would normally store the pid file for its master process.
+
+    GoogleString master_pid_filename = factory_->MasterPidFilename();
+    size_t final_slash = master_pid_filename.find_last_of('/');
+
+    return StrCat(
+        master_pid_filename.substr(0, final_slash), "/",
+        controller_pid_filename);
```
morlovich: This feels goofy if controller_pid_filename is something like foo/bar.
Which is a silly thing to set it too, but still.

pagespeed/system/controller_manager.cc
morlovich: +1 to not using strerror, actually, since it feels like much of the time I see the output the first step is grepping to figure out what code # it is anyway.

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,329 @@
...
+  }
+  handler_->Message(
+      kWarning, "Couldn't get controller process (%d) to exit", controller_pid);
+}
+
+void ControllerManager::RunController() {
+  handler_->Message(kInfo, "Controller running with PID %d", getpid());
+
+  // Would set up gRPC server here and pass control to its event loop.  Instead
+  // just hang out sleeping forever.
```
morlovich: The exit is handled by the watcher thread, right? May be worth mentioning that here.

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,329 @@
...
+  // If we disconnect file descriptors then logging will break, so don't.
+}
+
+void ControllerManager::ForkOffControllerProcess() {
+  handler_->Message(kInfo, "Forking controller process off of %d", getpid());
+
+  pid_t pid = fork();
+  CHECK(pid != -1) << "Couldn't fork a controller babysitter process";
+
+  if (pid != 0) return;  // Parent process.
```
morlovich: No 1-line ifs for us...

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,329 @@
...
+                                        babysitter_pid_));
+      CHECK(process_death_watcher_thread_->Start());
+
+      RunController();
+    } else {
+      // Wait for controller process to die, then continue with the loop by
+      // restarting it.
+      int status;
+      pid_t child_pid = waitpid(pid, &status, 0);
+      CHECK(child_pid != -1) << "Call to waitpid failed.";
```
morlovich: http://pubs.opengroup.org/onlinepubs/7908799/xsh/wait.html suggests EINTR is a possibility, even if Linux's version only seems to issue it in weird corner case...

morlovich:
The hard thing to see here is whether there are any races...

pagespeed/system/controller_manager.cc
jeffkaufman: added a check for this

net/instaweb/rewriter/public/rewrite_options.h
jeffkaufman: whoops; fixed.

pagespeed/apache/apache_rewrite_driver_factory.cc
jeffkaufman: It's unfortunate that we used "root" when apache and nginx use "master".  But I'll switch to "root" since the existing code uses that.

pagespeed/apache/apache_rewrite_driver_factory.cc
jeffkaufman: added comment

pagespeed/apache/mock_apache.cc
jeffkaufman: fixed

pagespeed/system/controller_manager.cc
jeffkaufman: You can't wait on a random pid, and because the root process daemonizes, it might as well be random.

We might be able to do something more efficient with pipes, where I keep open pipes between these, and then if the other end dies I'll see something happen.  I'll look into that.

pagespeed/system/controller_manager.cc
jeffkaufman: done

pagespeed/system/controller_manager.cc
jeffkaufman: added comment

pagespeed/system/controller_manager.cc
jeffkaufman: whoops, yes, that should be dynamic_cast

pagespeed/system/controller_manager.cc
jeffkaufman: added comment

pagespeed/system/controller_manager.cc
jeffkaufman: Oh?  The style guide is happy with this case: https://google.github.io/styleguide/cppguide.html#Conditionals

pagespeed/system/controller_manager.cc
jeffkaufman: made us loop on EINTR

pagespeed/system/controller_manager.h
jeffkaufman: whoops; that didn't belong there at all

pagespeed/system/controller_manager.h
jeffkaufman, line 67: no reason, moved to private

pagespeed/system/controller_manager.h
jeffkaufman: clarified comment

pagespeed/system/controller_manager.h
jeffkaufman: Yes, sorry, cleared this up some.   It's the babysitter pid in our code, and the "controller babysitter pid" to users.

pagespeed/system/system_rewrite_options.cc
jeffkaufman: clarified

jeffkaufman:
I think instead of looping watching to see if the master process pid has gone away I can use pipes.  I'll open a pipe in the master process, and close it for reading everywhere except the babysitter and controller where I'll close it for writing.  Then the controller can try to read from it, nonblocking:

> When attempting to read from an empty pipe or FIFO:
> - If no process has the pipe open for writing, `read()` shall return 0 to indicate end-of-file.
> - If some process has the pipe open for writing and `O_NONBLOCK` is set, `read()` shall return -1 and set errno to `EAGAIN`.

-- http://pubs.opengroup.org/onlinepubs/009695399/functions/read.html

This means that if the master (and all its children) are dead then we'll see EOF, otherwise we'll see -1.

I can watch two processes with `select` or something.

jeffkaufman:
Ok: heavily rewritten to use pipes instead of pid watching.  Pretty happy with this change.

pagespeed/apache/mock_apache.cc
```diff
@@ -160,6 +164,13 @@ ap_filter_rec_t* ap_register_input_filter(const char*, ap_in_filter_func,
   return NULL;
 }
 
+const char* ap_server_root_relative(apr_pool_t*, const char*) {
+  log_fatal("ap_server_root_relative");
+  return NULL;
+}
+
+const char* ap_pid_fname = NULL;
+
```
morlovich: I think these two can be removed now?

jeffkaufman: yes; done

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,228 @@
...
+  // If we disconnect file descriptors then logging will break, so don't.
+}
+
+void ControllerManager::ForkOffControllerProcess() {
+  handler_->Message(kInfo, "Forking controller process off of %d", getpid());
+
+  // Whenever we fork off a controller we save the fd for a pipe to it.  Then if
+  // we fork off another controller we can write a byte to the pipe to tell the
+  // old controller to clean up and exit.
+  static int controller_write_fd = -1;
```
morlovich, line 117: This might be easier to read if this is made class-member static, I think --- makes it obvious that there is this important global bit of state.

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,228 @@
...
+  if (controller_write_fd != -1) {
+    // We already forked off a controller earlier.  Tell it to quit by writing a
+    // byte.  If there's no one still with the pipe open we'll get SIGPIPE and
+    // die horribly, but the babysitter-restarting-controller structure makes
+    // that very unlikely.
```
morlovich: This doesn't sound very desirable, especially since aren't you in server master now?

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,228 @@
...
+        kInfo, "Writing a byte to a pipe to tell the old controller to exit.");
+    ssize_t status;
+    do {
+      status = write(controller_write_fd, "Q", 1);
+    } while (status == -1 && (errno == EAGAIN ||
+                              errno == EINTR));
+    if (status == -1) {
+      handler_->Message(kWarning, "killing old controller failed: %s",
+                        strerror(errno));
+    }
```
morlovich, line 133: Is a wait impossible here? This looks like it races between old controller dieing and new one starting (and potentially failing to grab the port because of that)?

pagespeed/system/controller_manager.cc
jeffkaufman: I agree it's not desirable.  But  as long as the babysitter hasn't died it won't happen.  The babysitter dies surprisingly if it can't `fork`, can't use `waitpid`, or if the controller calls `exit` without being asked to quit, and those shouldn't happen.

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,228 @@
...
+  // We need to fork again to make sure there is no session group leader.
+  pid_t pid = fork();
+  CHECK(pid != -1) << "Couldn't fork to daemonize.";
+  if (pid != 0) {
+    exit(EXIT_SUCCESS);
+  }
+
+  // If we keep the current directory we might keep them from being able to
+  // unmount their filesystem.
+  if (chdir("/") < 0) {
```
hillsp: Surely you only want to chdir in the child

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,228 @@
...
+}
+
+void ControllerManager::ForkOffControllerProcess() {
+  handler_->Message(kInfo, "Forking controller process off of %d", getpid());
+
+  // Whenever we fork off a controller we save the fd for a pipe to it.  Then if
+  // we fork off another controller we can write a byte to the pipe to tell the
+  // old controller to clean up and exit.
+  static int controller_write_fd = -1;
+  if (controller_write_fd != -1) {
```
hillsp, line 118: Can you move this code into a separate function?

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,228 @@
...
+  // just hang out sleeping until the ProcessDeathWatcherThread decides it's
+  // time to exit.
+  while (true) {
+    sleep(60);  // 1m
+  }
+}
+
+void ControllerManager::Daemonize() {
+  // Make a new session (process group).
+  if (setsid() < 0) {
```
hillsp, line 87: Shouldn't this be after you fork?

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,228 @@
...
+    return;
+  }
+
+  // Now we're in the child process.  Set this up as a babysitter process,
+  // that forks off a controller and restarts it if it dies.
+
+  Daemonize();
+  SetUpSignalHandlers();
+
+  factory_->set_is_root_process(false);
```
hillsp: This interface where you pass in the SystemRewriteDriver factory is a bit klunky. It seems like there should actually be a custom interface for this, though maybe that's too much abstraction.

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,228 @@
...
+  // EOF from the pipe it means we should quit because the master process shut
+  // down.
+  close(file_descriptors[1]);
+  int controller_read_fd = file_descriptors[0];
+
+  handler_->Message(kInfo, "Babysitter running with PID %d", getpid());
+
+  while (true) {
+    pid = fork();
+    CHECK(pid != -1) << "Couldn't fork a controller process";
```
hillsp, line 184: This is definitely something you could recover from. Are you worried about overloading the system if fork is failing?

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,228 @@
...
+      factory_->PrepareForkedProcess("controller");
+      factory_->PrepareControllerProcess();
+
+      // Start a thread to watch to see if the root or babysitter process
+      // dies, and quit if it does.
+      process_death_watcher_thread_.reset(new ProcessDeathWatcherThread(
+          thread_system_, controller_read_fd, handler_));
+      CHECK(process_death_watcher_thread_->Start());
+
+      RunController();
```
hillsp: explicitly call abort() here? Or kill(getpid()) ? Or even exit, for that matter, but don't fall back into the loop

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,228 @@
...
+    } while (status == -1 && (errno == EAGAIN ||
+                              errno == EINTR));
+    if (status == -1) {
+      handler_->Message(kWarning, "killing old controller failed: %s",
+                        strerror(errno));
+    }
+  }
+
+  int file_descriptors[2];
+  int pipe_status = pipe(file_descriptors);
```
hillsp, line 137: Why are you doing all this business with the pipe and not just killing the child when you want it to die?

pagespeed/system/controller_manager.cc
hillsp, line 117: I don't actually see why this actually needs to be static at all. Surely it's theoretically valid at least to have multiple babysitters, etc? (I know the port coming later actually steps on that)

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,228 @@
...
+    SystemRewriteDriverFactory* factory,
+    ThreadSystem* thread_system,
+    MessageHandler* handler) : factory_(factory),
+                               thread_system_(thread_system),
+                               handler_(handler) { }
+
+void ControllerManager::SetUpSignalHandlers() {
+  // We need to clear inherited signal handlers. There's no portable way to get
+  // a list of all possible signals, and they're not even guaranteed to be in
+  // order.  But NSIG is usually defined these days, and if it is then we just
```
jmarantz: what if it isn't defined?  Should there be an #ifndef?   Or is it better to fail the compile?

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,228 @@
...
+  // want ascending numbers up to to NSIG.
+  for (int i = 0; i < NSIG; i++) {
+    signal(i, SIG_DFL);
```
jmarantz: Is this API supported on Windows?

pagespeed/system/controller_manager.cc
jeffkaufman, line 117: Converted ControllerManager to being entirely static.  We weren't actually using state for very much at all.

pagespeed/system/controller_manager.cc
jeffkaufman, line 133: I don't think there's any way we can wait to start the new controller until after the old controller is dead, unless we watch PIDs which I'd like to avoid.  But the new controller can just try repeatedly to open up on that port, and once it succeeds the old controller is gone.

Added a comment to RunController to saying that the old controller may still be shutting down.

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,228 @@
...
+  ssize_t status;
+  char buf[1];
+  do {
+    status = read(controller_read_fd_, buf, 1);
+  } while (status == -1 && (errno == EINTR ||
+                            errno == EAGAIN));
+  if (status == -1) {
+    handler_->Message(
+        kWarning, "Controller got error %d reading from pipe, shutting down",
+        static_cast<int>(status));
```
jmarantz: also show errno & strerror(errno) ?

pagespeed/system/controller_manager.cc
jmarantz, line 100: nice forethought.

I wonder if you cd to /tmp instead, if any core-dumps might go there.  If you are in / they probably won't land.

pagespeed/system/controller_manager.cc
jeffkaufman, line 118: @hillsp Which code are you wanting moved?  Everything from the beginning of the function up to (and including) the `Daemonize(handler);` call?

pagespeed/system/controller_manager.cc
jeffkaufman, line 87: @hillsp We've already forked once at this point.  So now we're doing `setsid()` in a temporary process, and then forking again to make sure there's no session group leaded.  See the next step?

pagespeed/system/controller_manager.cc
jeffkaufman, line 184: Looking over the error codes for fork it really looks to me like if it's failing we're not going to be able to do anything useful:

```
       EAGAIN fork() cannot allocate sufficient memory to  copy  the  parent's
              page tables and allocate a task structure for the child.

           -> We already assume we can always allocate memory, and die if we can't.

       EAGAIN It was not possible to create a new process because the caller's
              RLIMIT_NPROC resource limit was  encountered.   To  exceed  this
              limit,  the  process  must  have either the CAP_SYS_ADMIN or the
              CAP_SYS_RESOURCE capability.

           -> We should not be hitting this limit by creating one new process, but I guess
               we could handle this case.  I'm not sure it's worth it, though.

       ENOMEM fork()  failed  to  allocate  the  necessary  kernel  structures
              because memory is tight.

           -> Same as EAGAIN          

       ENOSYS fork()  is not supported on this platform (for example, hardware
              without a Memory-Management Unit).

           -> Shouldn't apply.
```

pagespeed/system/controller_manager.cc
jeffkaufman, line 137: - We don't know the child's PID (it forks itself) so we'd need to look that up.
- We're already blocking `read`ing from the pipe to handle the case when the master goes away, so it's nice to use the same interface and not worry about signals.

pagespeed/system/controller_manager.cc
jeffkaufman: I'd rather just fail to compile, and add a workaround once we know what's necessary for the specific system that doesn't have it.  For example, it looks like on some platforms you have _NSIG instead, but we don't currently build on any of those.

pagespeed/system/controller_manager.cc
jeffkaufman: @jmarantz I think so: https://msdn.microsoft.com/en-us/library/xdkz3x12.aspx

@oschaaf Is there anything I'm doing here with forks, signals, and pipes that I should modify to let you support windows better?  Or hooks you'd like me to add, so you can patch out bits?

pagespeed/system/controller_manager.cc
jeffkaufman: @hillsp Actually, we don't need this. `PrepareForkedProcess` should set `is_root_process` to false.  Done.

This means we're using the factory for three things:
- `PrepareForkedProcess`
- `PrepareControllerProcess`
- passing it along to the controller, which probably needs it.

pagespeed/system/controller_manager.cc
jeffkaufman: Because `RunController` should never return?

Added LOG(FATAL).

pagespeed/system/system_rewrite_driver_factory.cc
```diff
@@ -239,6 +242,29 @@ void SystemRewriteDriverFactory::ParentOrChildInit() {
   SharedCircularBufferInit(is_root_process_);
 }
 
+void SystemRewriteDriverFactory::NameProcess(const char* name) {
+  // Set the process status.  This is what /proc/PID/status shows and what
+  // "ps -a" gives you.  With PR_SET_NAME there's a max of 16 characters, so
+  // abbreviate pagespeed as ps to be terse.
+  char name_for_prctl[16];
+  snprintf(name_for_prctl, sizeof(name_for_prctl), "ps-%s", name);
```
jmarantz, line 250: Wait...how about StrCat("ps-", StringPiece(name).substr(0, 13))

pagespeed/system/controller_manager.cc
jeffkaufman: Sorry, you're right, that should be `errno` instead of `status`.

`strerror` isn't thread safe, and `strerror_r` differs between gnu and posix in an annoying way, so I'd rather not deal with stringifying.

pagespeed/system/system_rewrite_driver_factory.cc
jeffkaufman, line 250: Why?

pagespeed/system/system_rewrite_driver_factory.h
```diff
@@ -258,6 +258,18 @@ class SystemRewriteDriverFactory : public RewriteDriverFactory {
     return 1;
   }
 
+  // Set the name of this process, for debugging visibility.
+  virtual void NameProcess(const char* name);
```
jmarantz, line 269: It seems unfortunate to have these process-management methods in the factory class.  Can't we just have the factory generate a process-manager that will have these methods?

pagespeed/system/controller_manager.h
```diff
@@ -0,0 +1,101 @@
...
+  // Takes ownership of nothing.
+  ControllerManager(SystemRewriteDriverFactory* factory,
+                    ThreadSystem* thread_system,
+                    MessageHandler* handler);
+  virtual ~ControllerManager() {}
+
+  // Called on system startup, before forking off any workers.  Starts up a
+  // babysitter process that starts a controller process and restarts the
+  // controller if it dies.  Also called (again) on configuration reloading.
+  void ForkOffControllerProcess();
```
jmarantz: I'm surprised this isn't virtual.

Let's think about how this will work on Windows, where we probably want a controller _thread_, no?

pagespeed/system/system_test.sh
```diff
@@ -2292,3 +2292,60 @@ start_test long url handling
 # This is an extremely long url, enough that it should give a 4xx server error.
 OUT=$($CURL -sS -D- "$TEST_ROOT/$(head -c 10000 < /dev/zero | tr '\0' 'a')")
 check_from "$OUT" grep -q "414 Request-URI Too Large\|Long"
+
+start_test babysitter process restarts controller when killed
+
+echo "$ERROR_LOG"
```
jmarantz: this is for debug?  delete?

pagespeed/system/system_rewrite_driver_factory.h
jeffkaufman, line 269: Right now the factory is where we have process global things, like `InitApr` and `IsServerThreaded`.  I'm adding `NameProcess`, `PrepareForkedProcess`, and `PrepareControllerProcess` which are all pretty small.

We could add a process manager to handle these, but I don't know if it's worth it.  And `PrepareControllerProcess` makes some `SystemRewriteDriverFactory`-internal setup calls.

pagespeed/system/controller_manager.h
jeffkaufman: Per offline discussion, I'm going to move the call for `ForkControllerProcess` into a virtual `SystemRewriteDriverFactory` method `StartController` so people can choose to do something else to provide the controller (like starting a thread).

jeffkaufman: My current expectation is that windows won't use the `ControllerManager` at all, but will either run the controller in a thread or will make controller calls directly ( @hillsp ? )

pagespeed/system/system_test.sh
jeffkaufman: Done

pagespeed/system/controller_manager.cc
hillsp, line 118: I was wondering if it made sense to move all the code that kills the child into a separate function, just for readability.

jeffkaufman:
Are there more comments?

morlovich:
Probably need to re-read this, since I don't think anyone looked at it for
last couple of weeks.

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,225 @@
...
+void ControllerManager::ProcessDeathWatcherThread::Run() {
+  handler_->Message(kInfo, "Watching the root process to exit if it does.");
+
+  ssize_t status;
+  char buf[1];
+  do {
+    status = read(controller_read_fd_, buf, 1);
+  } while (status == -1 && (errno == EINTR ||
+                            errno == EAGAIN));
+  if (status == -1) {
```
hillsp, line 51: Not sure all these cases are really worth worrying about. Seems like we you just always log "lost contact with controller, exiting" (maybe with the value of status) and leave it at that.

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,225 @@
...
+  // just hang out sleeping until the ProcessDeathWatcherThread decides it's
+  // time to exit.
+  while (true) {
+    sleep(60);  // 1m
+  }
+}
+
+void ControllerManager::Daemonize(MessageHandler* handler) {
+  // Make a new session (process group).
+  if (setsid() < 0) {
```
hillsp, line 87: Shouldn't this happen ~~before~~ after the fork()?

pagespeed/system/controller_manager.cc
jeffkaufman: Per separate discussion, the plan now is that ControllerManager won't be used on Windows/IIS.  In any non-forking system, you should just start a new thread that runs `ControllerManager::RunController`.  Added a comment for this.

pagespeed/system/controller_manager.cc
jeffkaufman, line 87: We fork twice, and this happens between the two forks.

First we fork to get a process we can mess with.  We call setsid in that process to detach from the previous process and be a session group leader.  Then we fork again (and have the parent die) to make sure there's no session group leader.

pagespeed/system/controller_manager.cc
jeffkaufman, line 51: The difference between pipe error, root process exiting, and root process starting a new controller seems like it would be useful to us for debugging in their field if something is wrong.  I'd rather keep these?

If you don't like the verbosity I guess we could log "lost contact with root process, exiting (status=%d, errno=%d)" but I think the way I have it is more likely to make sense to someone trying to debug this later.

jeffkaufman:
Please take a look; I'd love to get this in soon.

pagespeed/system/controller_manager.cc
hillsp: Depending what we're doing in this thread, that's not necessarily the right thing to do. For the controller in particular, it would be better to use the non RPC variant than to deal with all the serialisation overhead.

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,225 @@
...
+    // happen.
+    handler->Message(
+        kInfo, "Writing a byte to a pipe to tell the old controller to exit.");
+    ssize_t status;
+    do {
+      status = write(controller_write_fd, "Q", 1);
+    } while (status == -1 && (errno == EAGAIN ||
+                              errno == EINTR));
+    if (status == -1) {
+      handler->Message(kWarning, "killing old controller failed: %s",
```
hillsp, line 131: This is a bit racey since it assumes the controller will die immediately. Are you expecting the controller to just die if it can't bind to the port and retry? That's not super nice.

third_party/closure_library
```diff
@@ -1 +1 @@
-Subproject commit 2012c5372fdb02ce9531cf4b2561b05e3ce2ab39
```
hillsp, line 1: I assume you just need to sync for these?

pagespeed/system/controller_manager.cc
jeffkaufman: Makes sense.

Either way, the plan for non-forking systems like IIS/Windows is that they won't use a controller manager process, so there are no issues around using `signal()`, `fork()`, and other unixisms here.

third_party/closure_library
jeffkaufman, line 1: Yes, these will be fixed up before I submit.

pagespeed/system/controller_manager.cc
jeffkaufman, line 131: When the new controller starts the old controller may still be shutting down and may still have the port open.  The new controller should loop trying to get the port.

pagespeed/system/controller_manager.h
```diff
@@ -0,0 +1,95 @@
...
+#include <sys/types.h>
+
+#include "pagespeed/kernel/base/basictypes.h"
+#include "pagespeed/kernel/base/file_system.h"
+#include "pagespeed/kernel/base/message_handler.h"
+#include "pagespeed/kernel/base/scoped_ptr.h"
+#include "pagespeed/kernel/base/string.h"
+#include "pagespeed/kernel/base/thread.h"
+#include "pagespeed/kernel/base/thread_system.h"
+#include "pagespeed/kernel/base/timer.h"
```
morlovich: Suspect some of these aren't needed.

pagespeed/system/controller_manager.cc
```diff
@@ -0,0 +1,225 @@
...
+namespace net_instaweb {
+
+
+ControllerManager::ProcessDeathWatcherThread::ProcessDeathWatcherThread(
+    ThreadSystem* thread_system,
+    int controller_read_fd,
+    MessageHandler* handler) : Thread(thread_system,
+                                      "process death watcher",
+                                      ThreadSystem::kDetached),
+                              handler_(handler),
```
morlovich: +2 indent here and on the next line.

pagespeed/system/system_rewrite_driver_factory.h
```diff
@@ -258,6 +258,23 @@ class SystemRewriteDriverFactory : public RewriteDriverFactory {
...
+  // sense or running the Controller in its own process doesn't make sense, this
+  // is a hook where you can start the controller in whatever way makes sense
+  // for your platform.
+  virtual void StartController();
+
+  // Set the name of this process, for debugging visibility.
+  virtual void NameProcess(const char* name);
+
+  // Hook for handling any process-specific initialization the host webserver
+  // might need when we manually fork off a process.
```
morlovich: Mention that the base version should be called by subclasses, and what it does?
(Seems needed due to is_root_process_ assignment)

pagespeed/system/system_rewrite_driver_factory.h
```diff
@@ -258,6 +258,23 @@ class SystemRewriteDriverFactory : public RewriteDriverFactory {
...
+  virtual void PrepareForkedProcess(const char* name);
+
+  // Set up the controller process, with init, handlers, etc.
```
morlovich: couldn't really understand this comment

pagespeed/system/controller_manager.h
jeffkaufman: Fixed.

pagespeed/system/system_rewrite_driver_factory.h
jeffkaufman: Done

pagespeed/system/system_rewrite_driver_factory.h
jeffkaufman: Rewritten

pagespeed/system/controller_manager.cc
jeffkaufman: Done

jeffkaufman:
any other comments?

jeffkaufman:
Squashed and submitted as https://github.com/pagespeed/mod_pagespeed/commit/9bf75235c6899777898c004c0e213e5a6aefb6a7

PR #1286, opened 2016-04-01
Add support for rewriting data-src atttribute on images

jeffkaufman:


crowell:
after rebasing against master this will probably pass?

https://github.com/pagespeed/mod_pagespeed/commit/a2d0afab87a2ed622443c856e2fe63f8be85e3a1

fixes the reason why this fails in travis.

jeffkaufman:
This went in as https://github.com/pagespeed/mod_pagespeed/commit/ea3b78c7b6028a0bcca213d493a9a1b6514848c4

PR #1303, opened 2016-05-10
Test purging of ipro-rewritten resources.

jmarantz:
adds test for https://github.com/pagespeed/ngx_pagespeed/issues/1192

See https://github.com/pagespeed/ngx_pagespeed/pull/1193 for a fix.

### DIFF
--- a/pagespeed/system/system_test.sh
+++ b/pagespeed/system/system_test.sh
@@ -2143,6 +2143,7 @@ function cache_purge_test() {
       echo $CURL --request PURGE --proxy $SECONDARY_HOSTNAME "$PURGE_URL"
       check $CURL --request PURGE --proxy $SECONDARY_HOSTNAME "$PURGE_URL"
     fi
+    echo ""
     if [ $statistics_enabled -eq "0" ]; then
       # Without statistics, we have no mechanism to transmit state-changes
       # from one Apache child process to another, and so each process must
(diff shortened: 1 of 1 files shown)


pagespeed/system/system_test.sh
```diff
@@ -2143,6 +2143,7 @@ function cache_purge_test() {
       echo $CURL --request PURGE --proxy $SECONDARY_HOSTNAME "$PURGE_URL"
       check $CURL --request PURGE --proxy $SECONDARY_HOSTNAME "$PURGE_URL"
     fi
+    echo ""
```
oschaaf, line 2146: Left in by accident?

oschaaf:
travis complaints seem unrelated (font api css inlining failure)
had one remark about what seems to be an unintended echo line, otherwise LGTM

pagespeed/system/system_test.sh
jmarantz, line 2146: actually, no, this was on purpose.  For years I've been bothered a little bit by this message in the system-test logs:
    Purge successfulsleep 6
this brilliant bit of scripting makes it
    Purge successful
    sleep 6

oschaaf, line 2146: Ah cool :-)

jmarantz:
fyi this test implementation needs to be cleaned up; it writes files into a source-controlled area and this causes problems in our test scripts for a few reasons.

jmarantz:
Closing this pull-request; I need to push this one out in the other direction.  This business of writing files in a source-controlled area was something we were doing before, but it causes races in our test environment that I need a more infrastructural fix for.

PR #1309, opened 2016-05-13
Use full paths in ln -s and in cp -r, rather than using the cd-comman…

jmarantz:
…d-chain.


jmarantz:
note: this was lgtm'd internally.

PR #1310, opened 2016-05-13
setup the doc root before running vm_system_tests.

jmarantz:
This will hopefully fix the travis builds.


crowell:
LGTM, the failed push one was a timeout, the cache purge test is passing now.

PR #1312, opened 2016-05-18
Always parse process config into RewriteDriverFactory::default_options,

morlovich:
and then sync it back to per-ServerContext options. Should fix the
regression with ImageMaxRewritesAtOnce (https://github.com/pagespeed/mod_pagespeed/issues/1305),
and give RewriteDriverFactory and all the ServerContexts a consistent view of their value.

### DIFF
--- a/pagespeed/apache/mod_instaweb.cc
+++ b/pagespeed/apache/mod_instaweb.cc
@@ -328,9 +328,9 @@ class ApacheProcessContext {
     return factory_.get();
   }
 
-  // Checks cmd to see if it's used in a vhost or conditional context, and
-  // if so, if that's an error or warning condition.
-  const char* CheckCommandForVhost(const cmd_parms* cmd);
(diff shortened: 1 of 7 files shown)


pagespeed/apache/mod_instaweb.cc
```diff
@@ -988,6 +988,18 @@ int pagespeed_post_config(apr_pool_t* pool, apr_pool_t* plog, apr_pool_t* ptemp,
       CHECK(server_context != NULL);
       server_contexts.push_back(server_context);
     }
+
+    // We also want propagate all the per-process options to each vhost. The
+    // normal merge in merge_server_config isn't enough since that merges the
+    // non-per process things from a dummy ServerContext corresponding to the
```
morlovich, line 994: Review note for self: remove "dummy", since it does serve traffic.

morlovich:
Going to export the perforce version..

PR #1336, opened 2016-07-04
Add option to make ProxyFetch forward incoming flushes

oschaaf:
ngx_pagespeed currently buffers when rewriting HTML. This adds an
option to make ProxyFetch follow up on any incoming Flush calls
by requesting a Flush on it's associated RewriteDriver. This option
is on by default, on parity with mod_pagespeed's behaviour.

Includes a gtest for ensuring ProxyFetch complies and system tests
to test for consistent default flushing behaviour accross modules.

Will be accompanied by a PR to ngx_pagespeed's repository.
Edit:
The nps pull is:
https://github.com/pagespeed/ngx_pagespeed/pull/1217

### DIFF
--- a/pagespeed/apache/mod_instaweb.cc
+++ b/pagespeed/apache/mod_instaweb.cc
@@ -131,6 +131,7 @@ const char kModPagespeedForbidFilters[] = "ModPagespeedForbidFilters";
 const char kModPagespeedForceCaching[] = "ModPagespeedForceCaching";
 const char kModPagespeedExperimentVariable[] = "ModPagespeedExperimentVariable";
 const char kModPagespeedExperimentSpec[] = "ModPagespeedExperimentSpec";
+const char kModPagespeedFollowFlushes[] = "ModPagespeedFollowFlushes";
 const char kModPagespeedGeneratedFilePrefix[] =
     "ModPagespeedGeneratedFilePrefix";
 const char kModPagespeedGlobalAdminDomains[] = "ModPagespeedGlobalAdminDomains";
--- a/pagespeed/automatic/proxy_fetch.cc
+++ b/pagespeed/automatic/proxy_fetch.cc
@@ -961,7 +961,7 @@ bool ProxyFetch::HandleFlush(MessageHandler* message_handler) {
     // in ExecuteQueued.  Note that this can re-order Flushes behind
     // pending text, and aggregate together multiple flushes received from
     // the network into one.
-    if (Options()->flush_html()) {
+    if (Options()->flush_html() || Options()->follow_flushes()) {
       ScopedMutex lock(mutex_.get());
       network_flush_outstanding_ = true;
       ScheduleQueueExecutionIfNeeded();
@@ -1043,16 +1043,21 @@ void ProxyFetch::ExecuteQueued() {
     ScopedMutex lock(mutex_.get());
     DCHECK(!waiting_for_flush_to_finish_);
 
-    // See if we should force a flush based on how much stuff has
-    // accumulated.
     size_t total = 0;
     size_t force_flush_chunk_count = 0;  // set only if force_flush is true.
-    for (size_t c = 0, n = text_queue_.size(); c < n; ++c) {
-      total += text_queue_[c]->length();
-      if (total >= buffer_limit) {
-        force_flush = true;
-        force_flush_chunk_count = c + 1;
-        break;
+    if (network_flush_outstanding_ && Options()->follow_flushes()) {
+      force_flush = true;
+      force_flush_chunk_count = text_queue_.size();
+    } else {
+      // See if we should force a flush based on how much stuff has
+      // accumulated.
+      for (size_t c = 0, n = text_queue_.size(); c < n; ++c) {
+        total += text_queue_[c]->length();
+        if (total >= buffer_limit) {
+          force_flush = true;
+          force_flush_chunk_count = c + 1;
+          break;
+        }
       }
     }
 
--- a/pagespeed/automatic/proxy_fetch_test.cc
+++ b/pagespeed/automatic/proxy_fetch_test.cc
@@ -39,6 +39,7 @@
 #include "pagespeed/kernel/http/http_names.h"
 #include "pagespeed/kernel/http/request_headers.h"
 #include "pagespeed/kernel/http/response_headers.h"
+#include "pagespeed/kernel/thread/mock_scheduler.h"
 #include "pagespeed/kernel/thread/thread_synchronizer.h"
 #include "pagespeed/kernel/thread/worker_test_base.h"
 #include "pagespeed/kernel/util/platform.h"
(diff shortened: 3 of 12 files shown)


morlovich:
Hmm, how does the integration test work with the timer-based auto-flusher? It seems like there is a potential for flakes there. 

(Also I hope that works properly in nginx? Though I suppose it's not that effective when local end is producing a lot of data --- it was mostly meant for proxy use anyway)

oschaaf:
I did not test the timer-based auto-flusher, but I also think it's not enabled by default, so I wouldn't expect the test from this PR to be affected.

To enable the timer-based flushing, you would need to configure as follows, right?:

```
pagespeed FlushHtml on;
```

I can't find `FlushHtml` in the current test configuration. As far as I can tell, there currently is no integration test for this option in ngx_pagespeed.

oschaaf:
With regard to the timer-based flushing, I think `FlushHtml` is not documented today. It's a nice option indeed for those using `ngx_pagespeed` in a proxy setup, and perhaps worth adding to the docs?

pagespeed/apache/mod_instaweb.cc
```diff
@@ -1789,6 +1790,8 @@ static const command_rec mod_pagespeed_filter_cmds[] = {
   APACHE_CONFIG_OPTION(kModPagespeedBlockingRewriteRefererUrls,
                        "wildcard_spec for referer urls which trigger blocking "
                        "rewrites"),
+  APACHE_CONFIG_OPTION(kModPagespeedFollowFlushes,
+      "Attempt to mirror incoming flushes for html streams in the output"),
```
morlovich: Better to just give the option a name when registering this, then no code in the file is needed for mod_pagespeed to pick it up.
(Also it won't be inconsistent on whether it's directory vs. global or whatever.)

morlovich: ... Though the description is likely to be confusing, since mod_pagespeed doesn't generally use ProxyFetch, and I am pretty sure InstawebContext does pass through all the Apache flushes unconditionally.

(Also this doesn't follow the style guide --- can't have one line with arguments indented and another at +4).

pagespeed/automatic/proxy_fetch.cc
```diff
@@ -1035,7 +1035,8 @@ void ProxyFetch::ExecuteQueued() {
   bool do_flush = false;
   bool do_finish = false;
   bool done_result = false;
-  bool force_flush = false;
+  bool force_flush = network_flush_outstanding_
+      && Options()->follow_flushes();
```
morlovich: This changes the behavior with flush_html.
Also a bit weird to set network_flush_outstanding after checking flush_html  (along with || follow_flushes()) and then just ignore it.

pagespeed/automatic/proxy_fetch_test.cc
```diff
@@ -310,6 +311,89 @@ TEST_F(ProxyFetchTest, TestInhibitParsing) {
   mock_proxy_fetch->Done(true);
 }
 
+namespace {
+
+class FlushLoggingStringAsyncFetch : public StringAsyncFetch {
+ public:
+  FlushLoggingStringAsyncFetch(const RequestContextPtr& request_context)
```
morlovich: explicit.

pagespeed/automatic/proxy_fetch_test.cc
```diff
@@ -310,6 +311,89 @@ TEST_F(ProxyFetchTest, TestInhibitParsing) {
...
+      : StringAsyncFetch(request_context) {}
+
+  virtual ~FlushLoggingStringAsyncFetch() {}
+
+  virtual bool HandleFlush(MessageHandler* handler) {
```
morlovich: bool HandleFlush(MessageHandler\* handler) override {
these days.

pagespeed/automatic/system_test_helpers.sh
```diff
@@ -825,3 +825,51 @@ function kill_listener_port {
   PORT="$2"
   kill -9 $(lsof -t -i TCP:${PORT} -s TCP:LISTEN -a -c "/^${CMDLINE}$/") || true
 }
+
+# Performs timed reads on the output from a command passed via $1. The stream
```
morlovich, line 829: Any reason to have this here and not just follow_flushes.sh?

oschaaf, line 829: re: Any reason to have this here and not just follow_flushes.sh?
ngx_pagespeed will also add a test which needs this

pagespeed/apache/mod_instaweb.cc
oschaaf: Removed

pagespeed/automatic/proxy_fetch.cc
oschaaf: Re: This changes the behavior with flush_html.
Also a bit weird to set network_flush_outstanding after checking flush_html (along with || follow_flushes()) and then just ignore it.

Rewritten that part, better?

pagespeed/automatic/proxy_fetch_test.cc
oschaaf: done

pagespeed/automatic/proxy_fetch_test.cc
oschaaf: done

pagespeed/automatic/system_test_helpers.sh
oschaaf, line 829: re: Any reason to have this here and not just follow_flushes.sh?
companion PR for nps: https://github.com/pagespeed/ngx_pagespeed/pull/1217
(updated the commit description to mention that too)

oschaaf:
@morlovich addressed your comments and force pushed, PTAL?

pagespeed/apache/mod_instaweb.cc
```diff
@@ -131,6 +131,7 @@ const char kModPagespeedForbidFilters[] = "ModPagespeedForbidFilters";
 const char kModPagespeedForceCaching[] = "ModPagespeedForceCaching";
 const char kModPagespeedExperimentVariable[] = "ModPagespeedExperimentVariable";
 const char kModPagespeedExperimentSpec[] = "ModPagespeedExperimentSpec";
+const char kModPagespeedFollowFlushes[] = "ModPagespeedFollowFlushes";
```
morlovich, line 134: Unused now.

morlovich:
Looks good, though I am not certain we need yet another flush knob. Shall I merge this in?

oschaaf:
@morlovich It would be great if you could merge it.. Will you take care of removing `kModPagespeedFollowFlushes` in that case?

morlovich:
Yeah.  (Will also need to emulate your PHP thing in google-specific tests, which may take a bit..)

morlovich:
Hmm, this actually fails in Apache for me ---- the PHP flushes don't seem to produce APR_BUCKET_IS_FLUSH buckets for me....

morlovich:
Looks like it can be made to work with mod_Fastcgid handling PHP via:
FcgidOutputBufferSize 0

As hinted by this mod_fcgid changelog entry:
"If FCGI_Fflush() is called in your application, please add "OutputBufferSize 0" in you 
 httpd.conf, which will not keep any CGI output in cache buffer."

Should be easy enough to add this into debug.conf.template guarded by an <IfModule>

(Though I am having a different test fail afterwards, seemingly unrelated, but investigating)

morlovich:
... And the "unrelated failure" was because some debug output I added was confusing the grep'ing of the debug log in that test.

PR #1341, opened 2016-07-06
bump GRPC to 0.15

crowell:
builds under Centos5 with the following CFLAGS exported

```
export CFLAGS="-DGPR_MANYLINUX1 -std=gnu99"
```

any opinion on whether to always build with `GPR_MANYLINUX`?

```
--std=gnu99
```

can probably be safely added to the cflags for the whole project though.

### DIFF
--- a/third_party/grpc/generate_grpc_gyp
+++ b/third_party/grpc/generate_grpc_gyp
@@ -13,10 +13,10 @@ function gypify_file_list {
 
 cd src/third_party/grpc
 
-GPR_DIRS='src/src/core/lib/profiling src/src/core/lib/support src/src/core/ext/census'
-CORE_EXCLUDE_DIRS='src/src/test/core/statistics src/src/test'
+GPR_DIRS='src/src/core/lib/profiling src/src/core/lib/support'
+CORE_EXCLUDE_DIRS='src/src/test/core/statistics src/src/test src/src/core/ext/census'
 
-gpr_sources=$(find $GPR_DIRS -type f -name '*.c' ! -name 'window_stats.c' | gypify_file_list)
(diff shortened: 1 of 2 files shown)


third_party/grpc/generate_grpc_gyp
```diff
@@ -13,10 +13,10 @@ function gypify_file_list {
 
 cd src/third_party/grpc
 
-GPR_DIRS='src/src/core/profiling src/src/core/support'
-CORE_EXCLUDE_DIRS='src/src/core/statistics src/src/test'
+GPR_DIRS='src/src/core/lib/profiling src/src/core/lib/support src/src/core/ext/census'
+CORE_EXCLUDE_DIRS='src/src/test/core/statistics src/src/test'
 
-gpr_sources=$(find $GPR_DIRS -type f -name '*.c' | gypify_file_list)
+gpr_sources=$(find $GPR_DIRS -type f -name '*.c' ! -name 'window_stats.c' | gypify_file_list)
```
hillsp: Dropping just that one file out scares me a bit. You can't just exclude all of core/ext/census? What's up with that file in particular?

crowell: it contains a definition that doesn't exist, I'll try excluding that entire dir.

hillsp:
I'm inclined to only use GPR_MANYLINUX on systems that require it. I think it'll disable some stuff we might actually prefer to use. Given that, is it possible to only set cflags on the manylinux platforms, since presumably the new systems are already defaulting to at least c99.

crowell:
we can always just do export CFLAGS="..." before building on old systems,
and this can also go into the documentation as well

crowell:
currently hangs on system tests, looking into it now.

crowell:
closing this, i'll push the proper one tomorrow

PR #1375, opened 2016-08-12
Separate option for specifying a critical images beacon url.

oschaaf:
There are multiple reasons why someone would like to change the beacon url, and one of them is to have timings originating from the add_instrumentation filter processed on another machine.

Unfortunately, pointing `ModPagespeedBeaconUrl` to another server will also send the beacons that contain data about critical images critical images to the other server, which probably is no good.

This change proposes a separate option for beaconing data about critical images, `ModPagespeedCriticalImagesBeaconUrl`

Note that this should go in together with a separate change in ngx_pagespeed to not break its system tests.
- edit, add the nps PR-:
  https://github.com/pagespeed/ngx_pagespeed/pull/1245


jmarantz (approved):

oschaaf:
This needed some patching up to work with current master. I'm going to close this and re-open from a different branch

PR #1376, opened 2016-08-15
Make Travis run unit tests with apache_test make target

yeputons:
I'm waiting till <a href="https://travis-ci.org/pagespeed/mod_pagespeed/builds/152483485">build</a> succeeds and then will manually check logs to ensure that memcached and redis tests were not skipped.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,10 +23,9 @@ install:
   - sudo dpkg -i out/Release/mod-pagespeed*.deb
 script:
   - cd ~/build_directory/src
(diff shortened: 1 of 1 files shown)


crowell:
LGTM, let's wait for travis to finish, then merge.

yeputons:
Oops, we do not have that target in that repository. Looks like I have to run scripts manually from `.travis.yml`, will do that after refactor which make it easier to run with both memcached and redis.

PR #1377, opened 2016-08-16
Remove DCHECK in css summarizer

oschaaf:
This removes a DCHECK which occasionally fails. There's
a TODO which I think is related to this DCHECK. I propose
we create a new issue to track looking into tuning the memory
lifetime the TODO mentions upon merging this.

Fixes https://github.com/pagespeed/ngx_pagespeed/issues/891


morlovich:
The check isn't about the comment (or rather the comment should go with the
Clear()).

canceled_summaries_ is supposed to get cleared in RenderDone, which should
always happen....

oschaaf:
@morlovich Ah, in that case, this does not make sense, I'll close this.

PR #1387, opened 2016-09-06
Make this buildable again after Chromium SVN turndown

morlovich:
Besides the changes made in trunk, this require cherry-picking Steve's update of proto (since that also used to come from Chromium SVN), and picking an older RE2 release since MPS stable doesn't require C++11.


morlovich:
Looks like travis config needs to be updated, too...

net/instaweb/rewriter/redirect_on_size_limit_filter_test.cc
```diff
@@ -204,7 +215,7 @@ TEST_F(RedirectOnSizeLimitFilterTest, TestEscapingAndFlush) {
       "<table><tr><td>blah</td></tr></table></html>");
   html_parse()->FinishParse();
 
-  EXPECT_STREQ(kOutput, output_);
+  EXPECT_TRUE((kOutput == output_) || (kOutputAlt == output_)) << output_;
```
hillsp, line 218: Wasn't there a newer version of this fix that doesn't require kOutputAlt?

morlovich, line 218: There was a separate CL that tweaked the test, but I didn't think it was
important to backport.

hillsp: I'd have been inclined to backport the fix that matches trunk, but OK.

hillsp:
lgtm

PR #1394, opened 2016-09-13
strip-subresource-hints: respect preserve with rel=preload, use href not src

jeffkaufman:
- With rel=preload the hint tells us what type of resource it is, and if
  urls have been preserved for that type we should not strip it.
- If the rel=preload type isn't image, script, or style we shouldn't
  strip it, because those are the only urls we change.
- The filter was originally written to use src= when it should have used
  href=, which meant it removed hints it shouldn't have.

This is a minimal change for backporting for branch 33.  For the master branch
I have a draft of a more complex change that does additional cleanups.

Fixes: https://github.com/pagespeed/mod_pagespeed/issues/1392
Fixes: https://github.com/pagespeed/mod_pagespeed/issues/1393

### DIFF
--- a/net/instaweb/rewriter/strip_subresource_hints_filter.cc
+++ b/net/instaweb/rewriter/strip_subresource_hints_filter.cc
@@ -32,32 +32,48 @@ namespace net_instaweb {
 StripSubresourceHintsFilter::StripSubresourceHintsFilter(RewriteDriver* driver)
     : driver_(driver),
       delete_element_(NULL),
-      remove_(false) {
+      remove_script_(false),
+      remove_style_(false),
+      remove_image_(false),
+      remove_any_(false) {
 }
 
 StripSubresourceHintsFilter::~StripSubresourceHintsFilter() { }
 
 void StripSubresourceHintsFilter::StartDocument() {
   const RewriteOptions *options = driver_->options();
-  remove_ = (driver_->can_modify_urls() &&
-             (!(options->css_preserve_urls() &&
-                options->image_preserve_urls() &&
-                options->js_preserve_urls())));
+  remove_script_ = driver_->can_modify_urls() && !options->js_preserve_urls();
+  remove_style_ = driver_->can_modify_urls() && !options->css_preserve_urls();
+  remove_image_ = driver_->can_modify_urls() && !options->image_preserve_urls();
+  remove_any_ = remove_script_ || remove_style_ || remove_image_;
   delete_element_ = NULL;
 }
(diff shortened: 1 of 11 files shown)


jeffkaufman:
Travis failure was for a nullptr that needed to be a NULL; force-pushed to fix it.

net/instaweb/rewriter/strip_subresource_hints_filter.cc
```diff
@@ -32,32 +32,48 @@ namespace net_instaweb {
...
+    //   <link rel=subresource href=...> regardless
+    //   <link rel=preload as=script href=...> unless preserve scripts
+    //   <link rel=preload as=style href=...>  unless preserve styles
+    //   <link rel=preload as=image href=...>  unless preserve images
+    //
+    // Don't strip other kinds of rel=preload hints, because we don't change
+    // their urls, so existing ones are still valid.
+    const char* rel_value;
+    const char* as_value;
+    if ((rel_value = element->AttributeValue(HtmlName::kRel)) != NULL &&
```
hillsp, line 66: I do NOT love this code, but it does look correct.

net/instaweb/rewriter/strip_subresource_hints_filter_test.cc
```diff
@@ -213,8 +233,10 @@ TEST_F(StripSubresourceHintsFilterTest, SingleResourceDisallow) {
...
 
-TEST_F(StripSubresourceHintsFilterTestPartialPreserve,
-        MultiResourcePreserveAll) {
+// Even if you turn on preserve images, we still strip all rel=subresource hints
+// because we don't know which ones are images.
+TEST_F(StripSubresourceHintsFilterTestPreserveImage,
+       MultiSubresourcePreserveImages) {
   const char *source =
     "<head>"
     "<link rel=\"subresource\" src=\"/dontdropme.gif\"/>"
```
hillsp: Shouldn't you turn all these src attributes into href? Otherwise your "preserve" tests are kinda worthless.

jeffkaufman: Whoops!  Broke the test when backporting it.  This wouldn't have passed.

jeffkaufman:
(testing this by running `build_release_platform.sh` against `jefftk-fix-hints`)

hillsp:
LGTM assuming the tests pass 👍

jeffkaufman:
Build succeeded with passing tests!

PR #1402, opened 2016-09-22
Add a dummy stringpiece_util.h file so that include doesn't break OS builds

jmarantz:


jmarantz:
I am going to put this in using a different mechanism.  closing.

PR #1408, opened 2016-09-29
Fix versions for branch 34

hillsp:


morlovich:
LGTM.
We also need to update trunk to minor=13

PR #1410, opened 2016-10-03
Mark all scripts as executable

hillsp:
Simply ran the command from build_release_platform.sh:
find src -name "*.sh" | xargs chmod +x


jeffkaufman:
LGTM

possibly we should add passing along executable bits as part of our open source exporting process

hillsp:
I believe that happens by default, but yeah, I plan to experiment with that.

PR #1414, opened 2016-10-12
Cleanup broken submodules for DEPS

hillsp:


jeffkaufman:
Why are there more entries in 15e2eab than d870d1d?  I expected that for each submodule we were removing there would be a new `.gitignore` entry to keep us from adding it back?  Or is it that most of these submodules were accidentally added and then only later added to `.gitignore`?

hillsp:
@jeffkaufman That's the only explanation that makes sense to me. If the directories were in .gitignore, they should not have been added to the repo. That could be forensic'd out of the history if you really wanted to be certain.

jeffkaufman:
LGTM

PR #1424, opened 2016-10-27
Fix handling of failed AsyncFetches.

jeffkaufman:
`AsyncFetch::HandleDone(success=false)` can't pass `success=false` to `HeadersComplete`.  Success being `false` means "we've encountered an error" but the headers may or may not have been updated to match.  For example, if the `CacheUrlAsyncFetcher` doesn't find something it will set `status=false` and set up `404` headers, but if `SendFallbackResponse` aborts a response for having the wrong content type then it sets `status=false` while leaving `200` headers.

We need to make sure that error headers are sent whenever `status=false`, and the only place we can make sure of that is in `AsyncFetch::Done` itself, because `HandleDone` is too late and `HandleHeadersComplete` can't see `status`.

This fixes a flake in `resource_content_type_html.sh` with ngx_pagespeed wherewhen the deadline alarm was hit, which mostly happens under valgrind, it would send out the headers for the request with no body.

I've verified the fix by making sure that the test passes if you have set `test_instant_fetch_rewrite_deadline=true`, and also that while we previously got 5 flakes in 133 runs under valgrind now we don't get any.

(I previously tried to fix this on the ngx_pagespeed end with https://github.com/pagespeed/ngx_pagespeed/pull/1307 , but that doesn't work consistently because ngx_pagespeed races to send out headers immediately after receiving `HeadersComplete`.)

### DIFF
--- a/net/instaweb/http/async_fetch.cc
+++ b/net/instaweb/http/async_fetch.cc
@@ -107,11 +107,23 @@ void AsyncFetch::HeadersComplete() {
 
 void AsyncFetch::Done(bool success) {
   if (!headers_complete_) {
-    if (!success &&
-        (response_headers()->status_code() == 0)) {
-      // Failing fetches might not set status codes but we expect
-      // successful ones to.
-      response_headers()->set_status_code(HttpStatus::kNotFound);
+    if (!success) {
+      if (response_headers()->status_code() == 0) {
+        // Failing fetches might not set status codes but we expect
+        // successful ones to.
+        response_headers()->set_status_code(HttpStatus::kNotFound);
+      } else if (response_headers()->status_code() == HttpStatus::kOK) {
+        // Our API here is not ideal when Done is called with success=false.
+        // The problem is that we need to call HeadersComplete before we call
+        // HandleDone, but there's no way to tell HeadersComplete that we're in
+        // a bad state and it should throw away anything it has and return an
+        // error to the visitor.  This was a problem with ngx_pagespeed, because
+        // it sends out the headers immediately on HeadersComplete.
+        //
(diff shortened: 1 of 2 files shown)


net/instaweb/http/async_fetch.cc
```diff
@@ -107,11 +107,23 @@ void AsyncFetch::HeadersComplete() {
...
+        // The problem is that we need to call HeadersComplete before we call
+        // HandleDone, but there's no way to tell HeadersComplete that we're in
+        // a bad state and it should throw away anything it has and return an
+        // error to the visitor.  This was a problem with ngx_pagespeed, because
+        // it sends out the headers immediately on HeadersComplete.
+        //
+        // Since this can't be handled by the callee, handle it here.
+        response_headers()->Clear();
+        response_headers()->set_status_code(HttpStatus::kInternalServerError);
+      }
```
jmarantz, line 126: maybe DCHECK if !success && headers_complete() && response_headers()->status_code() is a successful code

morlovich:
LGTM

net/instaweb/http/async_fetch.cc
jeffkaufman: I just tried adding `DCHECK(success || response_headers()->IsErrorStatus());` right before `HandleDone(success);` but the problem is `RecordingFetch::HandleHeadersComplete()` called `set_response_headers(NULL)` so we don't have the same response headers anymore.  The basic problem is that it's not always ok to even look at response headers after headers complete.

So no `DCHECK`, sorry!

jeffkaufman:
Merged as 654230d

PR #1425, opened 2016-11-04
System test fixes for CentOS

hillsp:


hillsp:
Landed as part of commit ba0851fba8ca1f84ef6276d1c62e663a900d3d4e

PR #1426, opened 2016-11-07
Add dependencies for building a dev apache

jeffkaufman:


.gitmodules
```diff
@@ -104,3 +104,10 @@
 [submodule "tools/gyp"]
 	path = tools/gyp
 	url = https://chromium.googlesource.com/external/gyp
+# For building a development apache.
```
hillsp, line 107: How much data are you adding here? I guess it's pretty minor compared to apache?

jeffkaufman:
```
$ du -hs httpd
40M
```

hillsp:
lgtm

PR #1433, opened 2016-11-15
Add redis as a submodule

jeffkaufman:
Our redis tests require a version of redis more recent than the 2.8.4 that people developing on Ubuntu 14 LTS will have by default.  This is because we want to test cluster support, which was added in 3.0.  Instead of requiring people to set up redis themselves, include it as a submodule so build_development_apache.sh can use it.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -111,3 +111,6 @@
 [submodule "third_party/mod_fcgid"]
 	path = third_party/mod_fcgid
 	url = https://github.com/pagespeed/mod_fcgid.git
+[submodule "third_party/redis"]
+	path = third_party/redis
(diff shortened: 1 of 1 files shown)


hillsp:
There's already code to download redis in install_src_package in install/shell_utils.sh. If you're going to add a submodule, you should switch that over.

You could also consider hooking this up (and I guess this should also apply to nghttp2) so that it's only downloaded when install_required_packages is called with --additional_test_packages

jeffkaufman:
happy to switch over `install/shell_utils.sh`, but that will need to be in a different change because this one is opensource-only

jeffkaufman:
never mind

PR #1435, opened 2016-11-18
amp: add tests, and support image rewriting

jeffkaufman:


net/instaweb/rewriter/responsive_image_filter_test.cc
```diff
@@ -100,6 +100,13 @@ class ResponsiveImageFilterTest : public RewriteTestBase {
     ValidateExpected("test_simple", input_html, output_html);
   }
 
+  void TestSimple(int width, int height, StringPiece filename,
```
jeffkaufman, line 103: Comment from @jmarantz:

Looking at this choice I think I'd just put the tag into all the call-sites.  There are not that many.  Rather than adding this new variant.

Issue #1432, opened 2016-11-15
Get packaged for debian


jeffkaufman:
Debian has an existing bug out asking for mod_pagespeed integration: [debian:602316](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=602316).

Their docs on how to get started as a package maintainer: [Debian New Maintainers' Guide](https://www.debian.org/doc/manuals/maint-guide/)

I think the process here would be something like:
1) Start with the tarball for 1.11.33.4 ([1.11.33.4-r0.tar.bz2](https://dl.google.com/dl/linux/mod-pagespeed/tar/beta/mod-pagespeed-beta-1.11.33.4-r0.tar.bz2))
2) Using the debian `quilt` patching workflow, modify it to build against debian packages instead of bundled dependencies. (See the list I posted at the bottom of [debian:602316](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=602316).)
3) When we have something that works, start talking to debian.  We need to find a [sponsor], since we're not debian maintainers ourselves.

(Everything seems pretty resolvable, except for the situation with serf.  Not sure what to do there; we should talk to someone at debian.  But we can start the process of getting a debian-compliant tarball without that.)

jeffkaufman:
Debian's "no new packages" date for their next release is right after 1/1, so if we want to get in for Debian 9 we need to move quickly.

morlovich:
Hmm, what distro should we develop on? Don't think 9-pre-releases are
available on GCE; wonder if a diff with 8 would be material or not.

jeffkaufman:
It looks like you set up debian 8, and then upgrade it to unstable: [how do I install unstable](https://wiki.debian.org/InstallFAQ#Q._How_do_I_install_.22unstable.22_.28.22sid.22.29.3F)

morlovich:
Made some progress --- getting something that probably loads our .so (assuming I actually have everything in patches and not directly in source tree) , and mostly links natively. 
Attaching the Debian dir sketch thus far --- still lots of things incomplete, but I feel uncomfortable when I don't have a version control system backing me up.

[debian_2.tar.gz](https://github.com/pagespeed/mod_pagespeed/files/610014/debian_2.tar.gz)

morlovich:
Will attach some more after I work around github not liking tar.xz. 
[...]

PR #1436, opened 2016-11-21
Redirect following capability

### LINKED ISSUE #262: Follow redirects and eliminate them for sub-resources
```
Consider:

<img src="http://example.com/foo.png">

where:

requesting http://example.com/foo.png returns a 301 redirect to 
http://www.example.com/foo.png which is the actual image.

I believe that right now we will not rewrite or cache-extend this image because 
Serf will fail when not getting the 200 OK response for 
http://example.com/foo.png .

This is a big opportunity because not only can we rewrite and cache-extend the 
image, but we could remove unnecessary redirects.
```

Original issue reported on code.google.com by `sligocki@google.com` on 8 Apr 2011 at 6:58

oschaaf:
This adds the capability of following redirects for resource fetches to PageSpeed.

Fixes https://github.com/pagespeed/mod_pagespeed/issues/262


morlovich:
(Too busy with #1432, which suddenly got complicated to comment atm)

net/instaweb/http/public/redirect_following_url_async_fetcher.h
```diff
@@ -0,0 +1,81 @@
...
+class Statistics;
+class ThreadSystem;
+
+typedef std::set<GoogleString> GoogleUrlSet;
+
+class RedirectFollowingUrlAsyncFetcher : public UrlAsyncFetcher {
+ public:
+  // Does not take ownership of 'fetcher'.
+  RedirectFollowingUrlAsyncFetcher(UrlAsyncFetcher* fetcher,
+                                   GoogleString context_url,
```
morlovich: const GoogleString&

Also please document what context_url is here.

What's the lifetime story on the options?

net/instaweb/http/public/redirect_following_url_async_fetcher.h
```diff
@@ -0,0 +1,81 @@
...
+                                   ThreadSystem* thread_system,
+                                   Statistics* statistics, int max_redirects,
```
morlovich: probably want max_redirects earlier on --- in params should be before output ones per style guide.

net/instaweb/http/public/redirect_following_url_async_fetcher.h
```diff
@@ -0,0 +1,81 @@
...
+
+namespace net_instaweb {
+
+class AsyncFetch;
+class MessageHandler;
+class RewriteOptions;
+class Statistics;
+class ThreadSystem;
+
+typedef std::set<GoogleString> GoogleUrlSet;
```
morlovich: This is kinda confusing, the name makes it sound like a set of GoogleUrl's...

net/instaweb/http/public/redirect_following_url_async_fetcher.h
```diff
@@ -0,0 +1,81 @@
...
+  // Does not take ownership of 'fetcher'.
+  RedirectFollowingUrlAsyncFetcher(UrlAsyncFetcher* fetcher,
+                                   GoogleString context_url,
+                                   ThreadSystem* thread_system,
+                                   Statistics* statistics, int max_redirects,
+                                   RewriteOptions* rewrite_options);
+
+  virtual ~RedirectFollowingUrlAsyncFetcher();
+
+  virtual bool SupportsHttps() const { return base_fetcher_->SupportsHttps(); }
```
morlovich: bool SupportsHttps() const override, etc.

net/instaweb/http/public/redirect_following_url_async_fetcher.h
```diff
@@ -0,0 +1,81 @@
...
+
+  virtual void Fetch(const GoogleString& url, MessageHandler* message_handler,
+                     AsyncFetch* fetch);
+
+  class RedirectFollowingFetch;
+  friend class RedirectFollowingFetch;
```
morlovich: Don't actually need to friend those, we do this a lot by mistake, though. 
The nested class name should probably be inside private: section, too.

net/instaweb/http/redirect_following_url_async_fetcher.cc
```diff
@@ -0,0 +1,294 @@
...
+class RedirectFollowingUrlAsyncFetcher::RedirectFollowingFetch
+    : public SharedAsyncFetch {
+ public:
+  RedirectFollowingFetch(
+      RedirectFollowingUrlAsyncFetcher* redirect_following_fetcher,
+      AsyncFetch* base_fetch, const GoogleString& url,
+      const GoogleString& context_url, MessageHandler* message_handler)
+      : SharedAsyncFetch(base_fetch),
+        redirect_following_fetcher_(redirect_following_fetcher),
+        recieved_redirect_status_code_(false),
```
morlovich: recieved -> received

net/instaweb/http/redirect_following_url_async_fetcher.cc
```diff
@@ -0,0 +1,294 @@
...
+ protected:
+  virtual bool HandleFlush(MessageHandler* message_handler) {
+    if (!recieved_redirect_status_code_) {
+      return SharedAsyncFetch::HandleFlush(message_handler);
+    }
+    return true;
+  }
+
+  virtual void HandleHeadersComplete() {
+    GoogleString h = response_headers()->ToString();
```
morlovich: unused, presume was debug?

net/instaweb/http/redirect_following_url_async_fetcher.cc
```diff
@@ -0,0 +1,294 @@
...
+    if (my_success) {
+      // TODO(oschaaf): does this have the correct arguments?
+      bool cacheable = response_headers()->IsProxyCacheable(
+          base_fetch_->request_headers()->GetProperties(),
+          ResponseHeaders::GetVaryOption(options->respect_vary()),
+          ResponseHeaders::kHasValidator);
+      if (!cacheable) {
+        LOG(WARNING) << "Not following uncacheable redirect: " << url_;
+        my_success = false;
+      }
```
morlovich: *** The important one ***

Are 302 (or for that matters 307) ever cacheable?

Anyway, it's not enough to just check cacheability, you have to combine it -- if the redirect is valid for 1 day and the destination for 1 year, we can only cache for 1 day.

net/instaweb/http/redirect_following_url_async_fetcher.cc
```diff
@@ -0,0 +1,294 @@
...
+    const RewriteOptions* options =
+        redirect_following_fetcher_->rewrite_options();
+
+    if (!recieved_redirect_status_code_) {
+      SharedAsyncFetch::HandleDone(success);
+      delete this;
+      return;
+    }
+
+    bool my_success = success;
```
morlovich: can just change success

net/instaweb/http/redirect_following_url_async_fetcher.cc
```diff
@@ -0,0 +1,294 @@
...
+    int redirects_followed = urls_seen_->size();
+
+    if (my_success) {
+      // TODO(oschaaf): does this have the correct arguments?
+      bool cacheable = response_headers()->IsProxyCacheable(
+          base_fetch_->request_headers()->GetProperties(),
+          ResponseHeaders::GetVaryOption(options->respect_vary()),
+          ResponseHeaders::kHasValidator);
```
morlovich: Need to double-check kHasValidator in particular.

net/instaweb/http/redirect_following_url_async_fetcher.cc
```diff
@@ -0,0 +1,294 @@
...
+
+    if (!raw_redirect_url) {
+      LOG(WARNING) << "Failed looking up exactly one Location: header " << url_;
+      my_success = false;
+    } else if (!strlen(raw_redirect_url)) {
+      LOG(WARNING) << "Empty Location: header value";
+      my_success = false;
+    } else {
+      redirect_url = GoogleUrl::Sanitize(raw_redirect_url);
+      raw_redirect_url = redirect_url.c_str();
```
morlovich: raw_redirect_url is unused after this, I think..

net/instaweb/http/redirect_following_url_async_fetcher.cc
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    if (my_success && FindIgnoreCase(redirect_url, "#") != StringPiece::npos) {
```
morlovich: Don't need case insensitivity for #, so can probably just use redirect_url.find('#') ?

net/instaweb/http/redirect_following_url_async_fetcher.cc
```diff
@@ -0,0 +1,294 @@
...
+
+    scoped_ptr<GoogleUrl> redirect_gurl;
+    if (my_success) {
+      redirect_gurl.reset(new GoogleUrl(gurl_, redirect_url));
+      if (!redirect_gurl->IsWebValid()) {
+        LOG(WARNING) << "Invalid or unsupported url in location header: "
+                     << redirect_url;
+        my_success = false;
+      } else {
+        redirect_url = redirect_gurl->spec_c_str();
```
morlovich: redirect_gurl->Spec().CopyToString(&redirect_url);

pagespeed/system/system_server_context.cc
```diff
@@ -321,6 +322,13 @@ void SystemServerContext::ApplySessionFetchers(
...
+  if (options->max_fetch_redirects() > 0) {
+    driver->SetSessionFetcher(new RedirectFollowingUrlAsyncFetcher(
+        driver->async_fetcher(), system_request->url().as_string(),
+        thread_system(), statistics(), options->max_fetch_redirects(),
+        options));
+  }
+
   if (driver->options()->num_custom_fetch_headers() > 0) {
     driver->SetSessionFetcher(new AddHeadersFetcher(driver->options(),
                                                     driver->async_fetcher()));
```
morlovich: Hmm, might want to add your thing later in the list compares to the AddHeadersFetcher if we want every step of the redirect to get those custom headers.

morlovich (commented):
The big thing is that I don't think this compute final cacheability right?

net/instaweb/http/redirect_following_url_async_fetcher.cc
oschaaf: @morlovich I have worked on combining cacheability. 
I need to add tests for this, and address the other comments.

As for 302/307 responses being cacheable: I think they are but only when the headers explicitly say so. That is the behaviour this pr implements.

Some references

https://support.cloudflare.com/hc/en-us/articles/200168326-Are-301-and-302-redirects-cached-by-CloudFlare-
https://community.akamai.com/thread/1284
https://aprescott.com/posts/http-redirects

morlovich: I don't see an update, but maybe I am confused?
Also  ServerContext::ApplyInputCacheControl may be of some use, after some refactoring to make it not take resources?

My reference re: 302 and 307 was RFC 7231: https://tools.ietf.org/html/rfc7231#section-6.4.3
"Since the redirection might be altered on occasion, the client ought to continue to use the effective request URI for future requests."
It's more than a little awkward that it's using "ought to" rather than "MUST" or "SHOULD", though...

I am not sure how to take CDN stuff --- they may be doing a bit of non-compliance by default for better user-experience (I think we have some knobs in RewriteOptions like that that PSS used to set, for likely the same reason).  I probably err too much towards pedantic over pragmatic sometimes, though.

pagespeed/system/system_server_context.cc
morlovich: And this comment might be wrong due to request headers sharing?

net/instaweb/http/public/redirect_following_url_async_fetcher.h
jmarantz: Maybe just use StringSet from kernel/base/string_util.h ?

net/instaweb/http/redirect_following_url_async_fetcher.cc
```diff
@@ -0,0 +1,294 @@
...
+      }
+    }
+
+    if (my_success) {
+      if (redirects_followed > redirect_following_fetcher_->max_redirects()) {
+        LOG(WARNING) << "Max redirects " << redirects_followed << " / "
```
jmarantz: Do we want to have all these warnings echoed in a 404 page, rather than just putting them the server log?

Also it'd be better to carry a message_context_ so they go in the right server log.  E.g. in HTTPD there are vhost-specific loglevels.

net/instaweb/http/redirect_following_url_async_fetcher.cc
```diff
@@ -0,0 +1,294 @@
...
+
+  virtual bool HandleWrite(const StringPiece& content,
+                           MessageHandler* handler) {
+    if (!recieved_redirect_status_code_) {
+      return SharedAsyncFetch::HandleWrite(content, handler);
+    }
+    return true;
+  }
+
+  virtual void HandleDone(bool success) {
```
jmarantz: I feel like this function would benefit from breaking up into smaller methods for readability.

jmarantz:
Hey Otto -- are you going to follow up on the comments on this PR?  It'd be great to get this done.

oschaaf:
@morlovich could you take another look at this?
Note that this adheres to the http spec with regard to treatment of temporary redirects, I will follow up with a change that optionizes that.
I think this passes CI tests, but somehow it seems stuck on "in progress" (though when I click through the details, it shows checks as passed).

net/instaweb/http/redirect_following_url_async_fetcher.cc
```diff
@@ -0,0 +1,333 @@
...
+    const RewriteOptions* options =
+        redirect_following_fetcher_->rewrite_options();
+    const DomainLawyer* domain_lawyer = options->domain_lawyer();
+
+    bool cacheable = response_headers()->IsProxyCacheable(
+        base_fetch_->request_headers()->GetProperties(),
+        ResponseHeaders::GetVaryOption(options->respect_vary()),
+        ResponseHeaders::kNoValidator);
+
+    if (!cacheable) {
```
morlovich, line 227: Hmm, this + max_age_ covers the major cases, but it may be worth checking what happens with 
the subtler stuff like must-revalidate (which we probably can't handle well, come to think of it)  and 
s-maxage...

net/instaweb/http/redirect_following_url_async_fetcher.cc
```diff
@@ -0,0 +1,333 @@
...
+
+    if (!options->IsAllowed(redirect_gurl.Spec())) {
+      EmitRedirectWarning(context_url_, redirect_url, "Rewriting disallowed");
+      return false;
+    }
+
+    GoogleString mapped_domain_name;
+    GoogleString host_header;
+    bool is_proxy;
+    bool mapped = domain_lawyer->MapOriginUrl(
```
morlovich: Hmm, use RewriteOptionsManager::PrepareRequest instead of all this?

net/instaweb/http/redirect_following_url_async_fetcher.cc
```diff
@@ -0,0 +1,333 @@
...
+    } else {
+      // Shouldn't happen
+      DCHECK(false);
+      EmitRedirectWarning(url_, redirect_url, "Invalid mapped url");
+      return false;
+    }
+    return true;
+  }
+
+  bool TryExtractRedirectUrlFromResponseHeaders(GoogleString& redirect_url) {
```
morlovich: GoogleString*

net/instaweb/http/redirect_following_url_async_fetcher.cc
```diff
@@ -0,0 +1,333 @@
...
+    if (!raw_redirect_url) {
+      EmitRedirectWarning(
+          url_, "none", "Failed looking up exactly one Location header");
+      return false;
+    } else if (!strlen(raw_redirect_url)) {
+      EmitRedirectWarning(
+          url_, "", "Location header has an empty value");
+      return false;
+    } else {
+      redirect_url = GoogleUrl::Sanitize(raw_redirect_url);
```
morlovich: Why this?

net/instaweb/http/redirect_following_url_async_fetcher.cc
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    if (redirect_url.find('#') != StringPiece::npos) {
+      EmitRedirectWarning(
+          url_, redirect_url, "Location url has a fragment, not following");
+      return false;
+    }
+
+    scoped_ptr<GoogleUrl> redirect_gurl;
+    redirect_gurl.reset(new GoogleUrl(gurl_, redirect_url));
```
morlovich: No need to heap-allocate this, just GoogleUrl redirect_gurl(gurl_, redirect_url) will do.

net/instaweb/http/redirect_following_url_async_fetcher.cc
```diff
@@ -0,0 +1,333 @@
...
+    const GoogleString& url, MessageHandler* message_handler, AsyncFetch* fetch,
+    StringSet* redirects_followed_earlier, int64 max_age) {
+  RedirectFollowingFetch* redirect_following_fetch =
+      new RedirectFollowingFetch(this, fetch, url, context_url_,
+                                 redirects_followed_earlier, message_handler,
+                                 max_age);
+
+  if (redirect_following_fetch->Validate()) {
+    base_fetcher_->Fetch(url, message_handler, redirect_following_fetch);
+
```
morlovich: remove blank line..

pagespeed/system/system_tests/redirect_following.sh
```diff
@@ -0,0 +1,37 @@
...
+    $CURL -o/dev/null -sS --write-out '%{http_code}\n' "$URL")
+check [ "$OUT" = "404" ]
+
+start_test max redirects is respected
+URL=http://redirecting-fetch-single-only.example.com/redir_to_test/
+URL+=styles/A.1.css.pagespeed.cf.0.css
+echo "$URL"
+OUT=$(http_proxy=$SECONDARY_HOSTNAME \
+    $CURL -o/dev/null -sS --write-out '%{http_code}\n' "$URL")
+check [ "$OUT" = "404" ]
```
morlovich, line 43: This doesn't really cover the cache lifetime composition or the error cases. Might be easier with a unit test, though, since you can just feed whatever data you want...

morlovich (commented):
Looks pretty close...

install/debug.conf.template
```diff
@@ -1609,6 +1609,37 @@ ModPagespeedMessagesDomains Allow localhost
 </VirtualHost>
 
 <VirtualHost localhost:@@APACHE_SECONDARY_PORT@@>
+  ServerName redirecting-fetch.example.com
+  DocumentRoot "@@APACHE_DOC_ROOT@@"
+  ModPagespeedFileCachePath "@@MOD_PAGESPEED_CACHE@@"
+  ModPagespeedDomain redirecting-fetch.example.com:@@APACHE_SECONDARY_PORT@@
+  ModPagespeedMapOriginDomain 127.0.0.1:@@APACHE_SECONDARY_PORT@@ \
+    redirecting-fetch.example.com:@@APACHE_SECONDARY_PORT@@
```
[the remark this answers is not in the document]
oschaaf: The incoming request will be for `redirecting-fetch.example.com`. We use something like `--proxy 127.0.0.1:@@APACHE_SECONDARY_PORT@@` in the tests.
I added the origin mapping here as an equivalent to that, to get the redirects to land where they should..

net/instaweb/http/public/redirect_following_url_async_fetcher.h
```diff
@@ -0,0 +1,84 @@
...
+  static const int64 kUnset;
+
+  // Does not take ownership of 'fetcher'.
+  // The context_url is needed for verifying that the url we are about to
+  // redirect to is authorized.
+  RedirectFollowingUrlAsyncFetcher(UrlAsyncFetcher* fetcher,
+                                   const GoogleString& context_url,
+                                   ThreadSystem* thread_system,
+                                   Statistics* statistics, int max_redirects,
+                                   RewriteOptions* rewrite_options);
```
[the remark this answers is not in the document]
oschaaf: Yes, `statistics` is for future use, I'll make a common on that.

net/instaweb/http/redirect_following_url_async_fetcher.cc
```diff
@@ -0,0 +1,333 @@
...
+        redirect_following_fetcher_(redirect_following_fetcher),
+        received_redirect_status_code_(false),
+        urls_seen_(new StringSet()),
+        url_(url),
+        gurl_(url),
+        base_fetch_(base_fetch),
+        context_url_(context_url),
+        message_handler_(message_handler),
+        max_age_(kUnset) {
+    GoogleString sanitized = GoogleUrl::Sanitize(url);
```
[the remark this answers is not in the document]
oschaaf: My goal with `GoogleUrl::Sanitize` was to get the urls in a comparable form for checking that we do not follow redirects we already followed. Reading up on `Spec()`, that seems pretty thorough indeed, and it will save some work here. I'll change this to use `Spec()`.

install/debug.conf.template
```diff
@@ -1628,6 +1629,27 @@ ModPagespeedMessagesDomains Allow localhost
...
+  ModPagespeedMapOriginDomain 127.0.0.1:@@APACHE_SECONDARY_PORT@@ \
+    redirecting-fetch-csp.example.com:@@APACHE_SECONDARY_PORT@@
+
+  Alias "/redir_to_test" "@@APACHE_DOC_ROOT@@/mod_pagespeed_example"
+  Redirect permanent /redir_to_test/styles/blue.css http://redirecting-fetch.example.com:@@APACHE_SECONDARY_PORT@@/mod_pagespeed_example/styles/yellow.css
+
+  #Header set Content-Security-Policy "style-src http://foo.com/"
+  Header set Content-Security-Policy "default-src https:"
```
oschaaf: @morlovich this CSP response header does not seem to get picked up in the scan filter [1]. When I add this directive into the html as a meta tag, this works correctly. Do you have any clue where I should start with figuring out why the response header is not available in the response headers at that point in time? 

[1] https://github.com/pagespeed/mod_pagespeed/blob/master/net/instaweb/rewriter/scan_filter.cc#L60

PR #1440, opened 2016-11-30
Switch travis.yml to new build system, clean it up

hillsp:


jeffkaufman:
LGTM

hillsp:
The build timed out in apache_debug_stats_logging_test which is right near the end. I'm going to land this and we can iterate if needed.

PR #1444, opened 2016-12-08
ngxpagespeed.com configuration: fix /install

jeffkaufman:
ngxpagespeed.com/install serves a redirect to build_ngx_pagespeed.sh on github.  Now that we develop on master and trunk-tracking is gone, fix it to point at master.

Already fixed (manually) on the live site.


hillsp:
lgtm

PR #1447, opened 2016-12-12
licensing: correct license comments

jeffkaufman:
* many files had no license comments at all
* some files had license comments suggesting that they weren't open source
  (like a terse "all rights reserved") when they actually are open source.
* all our files are licensed under apache and should be marked as such


cfredric:
LGTM

cfredric (approved):

PR #1448, opened 2016-12-13
travis: add -luuid so we can build with devel on travis

jeffkaufman:
This should fix the ngx_pagespeed build on travis, because the ngx_pagespeed build is using devel.

(This is actually something I had originally done as part of preparing 91662d08a, but because part of that change involved removing our exporting tools I hand-exported the change and missed this file.)

### DIFF
--- a/third_party/apr/apr.gyp
+++ b/third_party/apr/apr.gyp
@@ -81,6 +81,7 @@
                   'link_settings': {
                     'libraries': [
                       '-ldl',
+                      '-luuid',
                   ]},
                 }],
               ],


install/build_development_apache.sh
```diff
@@ -77,7 +77,7 @@ cd "$third_party/apr/src"
 ./buildconf --prefix=$TARGET
 ./configure --prefix=$TARGET
 make
-make install
+MAKEFLAGS="" make install
```
hillsp: This fails if -j is set in MAKEFLAGS? Surely a better solution would be to just invoke make with -j 1

jeffkaufman: done: 51704cd

jeffkaufman:
Switching to building ngx_pagespeed on trusty turned out to be easier; doing that instead.

hillsp:
I was expecting a corresponding change to a script which installs libuuid-dev. Does that already happen somewhere?

jeffkaufman:
> I was expecting a corresponding change to a script which installs libuuid-dev. Does that already happen somewhere? 

That one is already in `install/ubuntu/install_required_packages.sh`, added in 91662d0, and this change was supposed to be in 91662d0 as well but I missed it in exporting.

PR #1449, opened 2016-12-16
Import the script used to prepare tarball for experimental packages

morlovich:
for Debian for https://github.com/pagespeed/mod_pagespeed/issues/1432

This version targets branch 33 so it goes there.

### DIFF
--- a/devel/create_distro_tarball_debian.sh
+++ b/devel/create_distro_tarball_debian.sh
@@ -1,0 +1,195 @@
+#!/bin/sh
+#
+# Copyright 2011 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Author: morlovich@google.com (Maksim Orlovich)
+#
+# The usual mechanism used to develop mod_pagespeed and build binaries is based
+# on merging all dependencies into a single source tree.  This script enables a
+# standard  untar/configure/make flow that does not bundle external libraries.
+# It generates the tarball including the configure (or rather generate.sh)
+# script.
+#
+# This is a variant that's extra-lean, as it targets Debian, which has a
+# particularly extensive package repository. Its file list is based on
+# version 1.11.33.4 (which is partly why it's a separate script and not just
+# a flag).
+
+set -e  # exit script if any command returns an error
+set -u  # exit the script if any variable is uninitialized
+
+function usage {
+  echo -n "create_distro_tarball_debian.sh [ --checkout_dir dir ] tarball"
+  echo " [ --git_tag tag | source_tree ]"
+  echo "examples of git_tag  would be 'master' or '1.11.33.4'"
+  exit 1
+}
+
+# This outputs a little wrapper around gyp that calls it with appropriate -D
+# flag
+function config {
+  cat <<SCRIPT_END
+#!/bin/sh
+#
+# This script uses gyp to generate Makefiles for mod_pagespeed built against
+# the following system libraries:
+#   apr, aprutil, apache httpd headers, icu, libjpeg_turbo, libpng, zlib.
+#
+# Besides the -D use_system_libs=1 below, you may need to set (via -D var=value)
+# paths for some of these libraries via these variables:
+#   system_include_path_httpd, system_include_path_apr,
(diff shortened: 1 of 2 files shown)


devel/create_distro_tarball_debian.sh
```diff
@@ -0,0 +1,195 @@
...
+
+CHECKOUT_DIR=
+if [ "$1" = "--checkout_dir" ]; then
+  if [ -z $2 ]; then
+    usage
+  fi
+  CHECKOUT_DIR=$2
+  shift 2
+else
+  CHECKOUT_DIR=`mktemp -d`
```
jeffkaufman: Same deal as in the other review: `$(...)` please

devel/create_distro_tarball_debian.sh
```diff
@@ -0,0 +1,195 @@
...
+# network access). Luckily, we can provide the number via LASTCHANGE.in,
+# so we just compute it now, and save it there.
+cd src/
+./build/lastchange.sh $PWD > LASTCHANGE.in
+cd ..
+
+# It's tarball time!
+# Note that this is highly-version specific, and may require tweaks for
+# every release as its dependencies change. It's probably a good idea to
+# branch this script with every release.
```
jeffkaufman: Same comment as in the other review

devel/create_distro_tarball_debian.sh
```diff
@@ -0,0 +1,195 @@
...
+
+# Also create a little helper script that shows how to run gyp
+config > $DIR/generate.sh
+chmod +x $DIR/generate.sh
+
+# Normally, the build system runs build/lastchange.sh to figure out what
+# to put into the last portion of the version number. We are, however, going to
+# be getting rid of the .svn dirs, so that will not work (nor would it without
+# network access). Luckily, we can provide the number via LASTCHANGE.in,
+# so we just compute it now, and save it there.
```
jeffkaufman: this comment talks about svn, so is probably obsolete

devel/create_distro_tarball_debian.sh
```diff
@@ -0,0 +1,195 @@
...
+# on merging all dependencies into a single source tree.  This script enables a
+# standard  untar/configure/make flow that does not bundle external libraries.
+# It generates the tarball including the configure (or rather generate.sh)
+# script.
+#
+# This is a variant that's extra-lean, as it targets Debian, which has a
+# particularly extensive package repository. Its file list is based on
+# version 1.11.33.4 (which is partly why it's a separate script and not just
+# a flag).
+
```
jeffkaufman: I'm not convinced this should be a separate script.  This seems like it should be a flag, and the 1.11 vs 1.12 vs master differences should be handled by having different copies on different branches

jeffkaufman (changes requested):

devel/create_distro_tarball_debian.sh
morlovich: I agree in principle. The main reason it's not done is there isn't any way of testing the equivalent of this for branch 34 (since that requires a lot of work wrt to gRPC and the like), not is there a working version of the other script for branch 33 around (though that one may be worth doing and is fairly straightforward to test, should I? --- that would provide a sensible way of merging it back into 34, too).

jeffkaufman: > is there a working version of the other script for branch 33 around?

Can't we just pull that out of (google-internal) source control history?

(And we'll need it if we need to do a patch release for 1.11)

devel/create_distro_tarball_debian.sh
morlovich: Done

devel/create_distro_tarball_debian.sh
morlovich: It should just talk about .git instead, change it here,

devel/create_distro_tarball_debian.sh
morlovich: I am not sure the same thing would work given the whole git change thing. It would be easy to change this back into a working one, since it's mostly changing foo/foo.gyp back into foo. Should I?

devel/create_distro_tarball_debian.sh
```diff
@@ -0,0 +1,195 @@
...
+set -e  # exit script if any command returns an error
+set -u  # exit the script if any variable is uninitialized
+
+function usage {
+  echo -n "create_distro_tarball_debian.sh [ --checkout_dir dir ] tarball"
+  echo " [ --git_tag tag | source_tree ]"
+  echo "examples of git_tag  would be 'master' or '1.11.33.4'"
```
morlovich: Changed this here (well, will once I push), is this better?

devel/create_distro_tarball_debian.sh
morlovich: Done

morlovich (commented):
Also, is it OK if I focus on this one, get it into a good shape, and then merge in to the trunk/34 one?

morlovich:
PTAL --- now with a flag. When done with this, I'll update the trunk/34 version based on it.

jeffkaufman (approved):
LGTM

PR #1450, opened 2016-12-16
Export the script we use to make release source tarballs

morlovich:
(this was tweaked slightly after the 34 release)

### DIFF
--- a/devel/create_distro_tarball.sh
+++ b/devel/create_distro_tarball.sh
@@ -1,0 +1,187 @@
+#!/bin/sh
+#
+# Copyright 2011 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Author: morlovich@google.com (Maksim Orlovich)
+#
+# The usual mechanism used to develop mod_pagespeed and build binaries is based
+# on merging all dependencies into a single source tree.  This script enables a
+# standard  untar/configure/make flow that does not bundle external libraries.
+# It generates the tarball including the configure (or rather generate.sh)
+# script.
+
+set -e  # exit script if any command returns an error
+set -u  # exit the script if any variable is uninitialized
+
+function usage {
+  echo -n "create_distro_tarball.sh [ --checkout_dir dir ] tarball"
+  echo " [ --git_tag tag | source_tree ]"
+  echo "examples of git_tag  would be 'master' or '1.11.33.4'"
+  exit 1
+}
+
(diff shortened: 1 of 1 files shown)


devel/create_distro_tarball.sh
```diff
@@ -0,0 +1,187 @@
...
+# It generates the tarball including the configure (or rather generate.sh)
+# script.
+
+set -e  # exit script if any command returns an error
+set -u  # exit the script if any variable is uninitialized
+
+function usage {
+  echo -n "create_distro_tarball.sh [ --checkout_dir dir ] tarball"
+  echo " [ --git_tag tag | source_tree ]"
+  echo "examples of git_tag  would be 'master' or '1.11.33.4'"
```
jeffkaufman, line 31: Can you clarify that `git_tag` can be a branch, tag, or commit id?  Basically, anything you can pass to `checkout -b`.

devel/create_distro_tarball.sh
```diff
@@ -0,0 +1,187 @@
...
+
+CHECKOUT_DIR=
+if [ "$1" = "--checkout_dir" ]; then
+  if [ -z $2 ]; then
+    usage
+  fi
+  CHECKOUT_DIR=$2
+  shift 2
+else
+  CHECKOUT_DIR=`mktemp -d`
```
jeffkaufman, line 74: can we swap the backtick command invocations for `$(...)`, here and below?

devel/create_distro_tarball.sh
```diff
@@ -0,0 +1,187 @@
...
+# so we just compute it now, and save it there.
+cd src/
+./build/lastchange.sh $PWD > LASTCHANGE.in
+cd ..
+
+# It's tarball time!
+# TODO(morlovich): Fix giflib.gyp to support proper external depends.
+# Note that this is highly-version specific, and may require tweaks for
+# every release as its dependencies change. It's probably a good idea to
+# branch this script with every release.
```
jeffkaufman, line 133: Can you change this to be more explicit?  Something like:

```
# Note that this is highly-version specific, and requires tweaks for every
# release as its dependencies change.  Always run the version of this
# script that's on the branch you're making a tarball for.
```

devel/create_distro_tarball.sh
```diff
@@ -0,0 +1,187 @@
...
+tar cj --dereference --exclude='.git' --exclude='.svn' --exclude='.hg' -f $TARBALL \
```
jeffkaufman, line 134: `--exclude='.svn' --exclude='.hg'` shouldn't be needed anymore now that all our dependencies are git.

devel/create_distro_tarball.sh
```diff
@@ -0,0 +1,187 @@
...
+    $DIR/src/third_party/modp_b64 \
+    $DIR/src/third_party/optipng \
+    $DIR/src/third_party/protobuf \
+    $DIR/src/third_party/rdestl \
+    $DIR/src/third_party/re2 \
+    $DIR/src/third_party/redis-crc \
+    $DIR/src/third_party/serf \
+    $DIR/src/third_party/zlib/zlib.gyp \
+    $DIR/src/tools/gyp \
+    $DIR/src/url
```
jeffkaufman, line 183: Could we exclude the windows dlls that lintian was objecting to here as well?

jeffkaufman (changes requested):

morlovich:
I am going to re-do it starting with a merge off branch 33 one, I think.

PR #1451, opened 2016-12-16
devel: fix flakes in build_development_apache

jeffkaufman:
build_development_apache.sh flakes about 21% of the time, due to the apr
makefile not handling parallelism properly when running the install target.
Explicitly set -j1 when installing apr.

Fixes pagespeed/ngx_pagespeed#1338

### DIFF
--- a/install/build_development_apache.sh
+++ b/install/build_development_apache.sh
@@ -77,7 +77,7 @@ cd "$third_party/apr/src"
 ./buildconf --prefix=$TARGET
 ./configure --prefix=$TARGET
 make
-make install
+make install -j1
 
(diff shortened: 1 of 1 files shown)


install/build_development_apache.sh
```diff
@@ -77,7 +77,7 @@ cd "$third_party/apr/src"
 ./buildconf --prefix=$TARGET
 ./configure --prefix=$TARGET
 make
-make install
+make install -j1
```
hillsp, line 80: I was expecting make -j1 install. I guess this works though?

hillsp:
LGTM

PR #1453, opened 2016-12-19
run-on-vm: add support for running checkin tests on a vm

jeffkaufman:
Split the 'stand up a vm / delete it when done' portion of build_on_vm.sh
off into a new script, run_on_vm.sh, and add checkin_on_vm.sh that uses it
to run checkin tests.

### DIFF
--- a/devel/checkin_on_vm.sh
+++ b/devel/checkin_on_vm.sh
@@ -79,10 +79,10 @@ function machine_ready() {
   sudo apt-get -y upgrade
   sudo apt-get -y install git
   if ! [ -d mod_pagespeed ]; then
-    git clone -b "$branch" https://github.com/pagespeed/mod_pagespeed.git
+    git clone -b "$branch" https://github.com/pagespeed/mod_pagespeed.git \
+              --recurse-submodules
   fi
   cd mod_pagespeed
-  git submodule update --init --recursive
   if ! [ -d ~/apache2 ]; then
     install/build_development_apache.sh 2.2 prefork
   fi
--- a/devel/checkin_on_vm.sh
+++ b/devel/checkin_on_vm.sh
@@ -75,6 +75,8 @@ function machine_ready() {
   gcloud compute ssh "$machine_name" -- bash << EOF
   set -e
   set -x
+  set -u
+
   sudo apt-get -y update
   sudo apt-get -y upgrade
   sudo apt-get -y install git
--- a/devel/checkin_on_vm.sh
+++ b/devel/checkin_on_vm.sh
@@ -67,6 +67,7 @@ if [ -z "$machine_name" ]; then
   machine_name="${USER}-checkin-${sanitized_branch}"
 fi
 
+# Empty final argument is to placate -u
 remaining_arguments=( "$@" "" )
 
 # Hook for run_on_vm.sh to call.
--- a/devel/checkin_on_vm.sh
+++ b/devel/checkin_on_vm.sh
@@ -1,0 +1,92 @@
+#!/bin/bash
+#
+# Copyright 2016 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Author: jefftk@google.com (Jeff Kaufman)
+#
+# Runs mod_pagespeed checkin tests on a gcloud VM.
+#
+# You may want to set CLOUDSDK_COMPUTE_REGION and/or CLOUDSDK_CORE_PROJECT,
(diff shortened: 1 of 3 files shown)


install/run_on_vm.sh
```diff
@@ -0,0 +1,88 @@
...
+
+# Display an error including the machine name if we die in the script below.
+trap '[ $? -ne 0 ] && echo -e "\nScript failed on $machine_name"' EXIT
+
+echo "Checking whether $machine_name is up; you may see some errors..."
+while ! gcloud compute ssh "$machine_name" -- true; do
+  echo "Waiting for $machine_name to come up..."
+done
+
+machine_ready
```
hillsp, line 83: Hmm, given the way you've done this, what do you think about instead just using the trap on line 76 to delete the machine, then you don't have the awkward callback into machine_ready.

install/build_on_vm.sh
```diff
@@ -99,28 +94,13 @@ if [ -z "$machine_name" ]; then
...
-if [ -z "$instances" ] || ! $use_existing_machine; then
-  gcloud compute instances create "$machine_name" \
-    --image-family="$image_family" --image-project="$image_project"
-fi
-
 mkdir -p ~/release
 
-# Display an error including the machine name if we die in the script below.
-trap '[ $? -ne 0 ] && echo -e "\nBuild failed on $machine_name"' EXIT
+remaining_arguments=( "$@" "" )
```
hillsp, line 100: Why the trailing "" ? That doesn't look right

install/build_on_vm.sh
```diff
@@ -130,11 +110,10 @@ gcloud compute ssh "$machine_name" -- bash << EOF
   fi
   git clone -b "$branch" https://github.com/pagespeed/mod_pagespeed.git
   cd mod_pagespeed
-  install/build_release.sh $@
+  install/build_release.sh ${remaining_arguments[@]}
```
hillsp: Oh, was the trailing quotes on remaining_arguments to deal with set -u here?

Maks' solution for this was to test the size of the array first, a la:
if [ "${#remaining_arguments[@]}" -gt 0 ]; then
  install/build_release.sh ${remaining_arguments[@]}
else
  install/build_release.sh
fi

The fancier approach I used in centos/install_required_packages.sh is:
install/build_release.sh ${remaining_arguments[@]+"${remaining_arguments[@]}"}

(from http://stackoverflow.com/questions/7577052/bash-empty-array-expansion-with-set-u)

devel/checkin_on_vm.sh
```diff
@@ -0,0 +1,92 @@
...
+  sanitized_branch="$(tr _ - <<< "$branch" | tr -d .)"
+  machine_name="${USER}-checkin-${sanitized_branch}"
+fi
+
+remaining_arguments=( "$@" "" )
+
+# Hook for run_on_vm.sh to call.
+function machine_ready() {
+  gcloud compute ssh "$machine_name" -- bash << EOF
+  set -e
```
hillsp, line 76: Do you want a set -u here?

devel/checkin_on_vm.sh
```diff
@@ -0,0 +1,92 @@
...
+  set -x
+  sudo apt-get -y update
+  sudo apt-get -y upgrade
+  sudo apt-get -y install git
+  if ! [ -d mod_pagespeed ]; then
```
hillsp, line 83: Why not just git clone --recursive ?

devel/checkin_on_vm.sh
```diff
@@ -0,0 +1,92 @@
...
+    git clone -b "$branch" https://github.com/pagespeed/mod_pagespeed.git
+  fi
+  cd mod_pagespeed
+  git submodule update --init --recursive
+  if ! [ -d ~/apache2 ]; then
+    install/build_development_apache.sh 2.2 prefork
+  fi
+  devel/checkin ${remaining_arguments[@]}
```
hillsp: Quote remaining arguments

hillsp (changes requested):

install/build_on_vm.sh
jeffkaufman: added comment 9488d44

install/build_on_vm.sh
jeffkaufman: Just doing `( "$@" "" )` seems much tidier to me?

devel/checkin_on_vm.sh
jeffkaufman: done (and in build_on_vm.sh): bbdf299

devel/checkin_on_vm.sh
jeffkaufman: done: d8844a0

devel/checkin_on_vm.sh
jeffkaufman: done: 0176598

install/run_on_vm.sh
jeffkaufman: We don't want to delete the machine if the thing we run on it fails, because someone may want to ssh in and look at it.

jeffkaufman:
Ready for another round of review

install/run_on_vm.sh
hillsp: Take a look at how the trap works, you can test $? and in fact I already do.

install/build_on_vm.sh
hillsp: That's going to insert an empty last argument, which I don't think you want?

PR #1455, opened 2016-12-20
Tarball script version for master & 34

morlovich:


devel/create_distro_tarball.sh
```diff
@@ -0,0 +1,226 @@
...
+# be getting rid of the .git dirs, so that will not work (nor would it without
+# network access). Luckily, we can provide the number via LASTCHANGE.in,
+# so we just compute it now, and save it there.
+cd src/
+./build/lastchange.sh $PWD > LASTCHANGE.in
+cd ..
+
+# Things that depends on minimal or not.
+if [ $MINIMAL -eq 0 ]; then
+  GTEST=$DIR/src/testing
```
hillsp, line 144: wdyt about just having one variable with all this crap in, instead of multiple? Obviously there are trade-offs there

devel/create_distro_tarball.sh
```diff
@@ -0,0 +1,226 @@
...
+# the .gyp[i] files and doesn't work out of the box. The pruning was also done
+# as of branch 33, so further tweaks might be required for this mode in
+# 34 or newer.
+
+set -e  # exit script if any command returns an error
+set -u  # exit the script if any variable is uninitialized
+
+function usage {
+  echo -n "create_distro_tarball_debian.sh [ --minimal ] [ --checkout_dir dir ] tarball"
+  echo " [ --git_tag tag | source_tree ]"
```
hillsp: I think maybe I'd say '--checkout commit_ish' which is the git way of saying "something that can be checked out.

devel/create_distro_tarball.sh
```diff
@@ -0,0 +1,226 @@
...
+#   system_include_path_aprutil.
+#
+# for example, you might run
+# ./generate.sh -Dsystem_include_path_apr=/usr/include/apr-1 \\
+#               -Dsystem_include_path_httpd=/usr/include/httpd
+# to specify APR and Apache include directories.
+#
+# Also, BUILDTYPE=Release can be passed to make (the default is Debug).
+echo "Generating src/Makefile"
+src/build/gyp_chromium -D use_system_libs=1 \$*
```
hillsp, line 64: This should include --depth=. otherwise it wont work if your checkout isn't in a direcotry named 'src'. Apparently I missed that when I was messing with the build system.

devel/create_distro_tarball.sh
```diff
@@ -0,0 +1,226 @@
...
+SCRIPT_END
+}
+
+if [ $# -lt 2 ]; then
+  usage
+  exit
+fi
+
+CHECKOUT_DIR=
```
hillsp, line 73: This would be SO MUCH more readable if it used getopt. Take a look here:

https://github.com/pagespeed/mod_pagespeed/blob/d55c2d7a2c64fab183f77571a18dad80ffe3cb1b/install/build_on_vm.sh#L34

devel/create_distro_tarball.sh
```diff
@@ -0,0 +1,226 @@
...
+    usage
+  else
+    GIT_TAG=$3
+    if [ -z $CHECKOUT_DIR ]; then
+      CHECKOUT_DIR=$(mktemp -d)
+    fi
+    SRC_DIR=$CHECKOUT_DIR
+    cd $CHECKOUT_DIR
+    git clone https://github.com/pagespeed/mod_pagespeed.git src\
+      --branch "$GIT_TAG" --recursive
```
hillsp: Beware of this; It turns out that on CentOS 6, --branch <tag> will print a warning but silently checkout master.

devel/create_distro_tarball.sh
morlovich: Sure, but I feel like it's outside the scope of this.

devel/create_distro_tarball.sh
morlovich: I've strongly considered it, the reason I picked this was to make it easier to see the two sides next to each other.

devel/create_distro_tarball.sh
morlovich: I would find that less helpful.

devel/create_distro_tarball.sh
morlovich: Could you elaborate? A lot of stuff assumes src/ below.

devel/create_distro_tarball.sh
hillsp: OK, but what you have is objectively wrong since 34 isn't a tag.

morlovich: Fair enough. Done, as suggested. (Not pushed yet, waiting for further feedback)

devel/create_distro_tarball.sh
hillsp: Sorry, I wasn't reading carefully enough. It's fine as-is.

The background is that gyp looks for a directory containing "src" in the path to figure out the root (which it calls "depth", for some reason). If you build from a directory that isn't src, it doesn't work. The pattern we use everywhere else is:

git clone <some_dir>
cd <some_dir>
gyp --depth=.

That way we can run gyp in the root of the checkout and it doesn't matter what the checkout is called. Doing the same thing wouldn't be awful for consistency, but isn't required.

devel/create_distro_tarball.sh
hillsp: How about the fact that I'm struggling to review the code because I don't understand how the arg parsing works?

devel/create_distro_tarball.sh
hillsp: git clone [...] src
cd src
git checkout $GIT_TAG

devel/create_distro_tarball.sh
hillsp: Fair enough. No reason we can't change that later

hillsp:
There are MANY things wrong with this script, but it's not significantly worse than the original. Don't forget to update the release docs delete the old script once this lands.

LGTM

jeffkaufman (approved):
LGTM

(Similar feelings to @hillsp : this script is still a pain, but your changes make it better.)

PR #1457, opened 2016-12-20
Update the BoringSSL instructions to match current paths,

morlovich:
submodule update directions (thanks to Steve to help with that!)

### DIFF
--- a/third_party/serf/UPDATING_BORINGSSL
+++ b/third_party/serf/UPDATING_BORINGSSL
@@ -5,7 +5,21 @@ have to take the gypi file from chromium's stable and modify the paths.
 The latest stable version of chromium can be found here.
 http://googlechromereleases.blogspot.com/search/label/Stable%20updates
 
-  SERF_DIR=/path/to/client/google3/third_party/instaweb/src/third_party/serf
+Export the version number as $CHROMIUM_VERSION
+
+  To update BoringSSL to latest chromium-stable proper, and include the update
+  in a pending commit:
+    cd third_party/boringssl/src
+    git checkout chromium-stable
+    cd ..
+    git add src
+    cd ../../
+
+  Then:
+  SERF_DIR=$PWD/third_party/serf
+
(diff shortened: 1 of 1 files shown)


morlovich:
Jeff, this should re-run .gyp based on the .gypi changing automatically, right?

third_party/serf/UPDATING_BORINGSSL
```diff
@@ -26,6 +40,3 @@ http://googlechromereleases.blogspot.com/search/label/Stable%20updates
     cp -r src/third_party/boringssl/$arch $SERF_DIR/$arch/
   done
```
hillsp, line 42: Dunno if it's worth calling out that you need to cd back into the git directory, commit stuff and push? It's probably also a good idea to spin a build before pushing at least.

But this looks much better, thanks.

hillsp:
On the subject of gyp, I don't think we have any automatic rules that update makefiles with the gypfiles change. One just has to re-run gyp "manually"

jeffkaufman:
@morlovich wrote:
> this should re-run .gyp based on the .gypi changing automatically, right?

@hillsp wrote:
> I don't think we have any automatic rules that update makefiles with the gypfiles change. One just has to re-run gyp "manually"

All existing gyp(i) files are listed in `mod_pagespeed/gyp_manifest.d`, a temporary file that's created when you first start using `devel/Makefile` and loaded as `GYP_FILES`  Then if any of those existing gyp files change that invalidates `HOOKS_STAMP`, making `gyp` rerun itself:

```
$(HOOKS_STAMP) : $(GPERF) $(GYP_FILES)
        bash -o pipefail -c \
          "cd $(GIT_SRC) && python build/gyp_chromium --depth=. | \              
             egrep -v '^(Generating )[a-zA-Z0-9_/.-]*(Makefile|mk)$$'"
        touch $(HOOKS_STAMP)
```

jeffkaufman (approved):
LGTM

PR #1459, opened 2016-12-20
doc: open source the user-facing documentation

jeffkaufman:
This is an open source version of the user-facing docs on [developers.google.com/speed/pagespeed/module/](https://developers.google.com/speed/pagespeed/module/).  I currently have it staged at [modpagespeed.com/doc](https://modpagespeed.com/doc), which is probably where this should go when live.

It uses server-side includes for headers and footers, which isn't as nice as being fully static, but should be more maintainable.

This change doesn't include a landing page for mod_pagespeed.  I think that should probably go on modpagespeed.com / ngxpagespeed.com and not in the docs directly.  Having it in the docs is a weird artifact of how we were hosted on developers.google.com.  But this is not in this CL.

Mobile PSI score is 100 with mod_pagespeed and 89 without, so we're also doing a bit of demonstrating pagespeed.

For reviewability, this is split into three sequential changes:
* https://github.com/pagespeed/mod_pagespeed/pull/1464 : straight copy-over of existing docs
* https://github.com/pagespeed/mod_pagespeed/pull/1465 : formatting changes that apply in bulk to all html files
* https://github.com/pagespeed/mod_pagespeed/pull/1459 (this change) : css and js to make the docs be usable and nice to look at

### DIFF
--- a/doc/.htaccess
+++ b/doc/.htaccess
@@ -1,0 +1,14 @@
+# Turn on server side include processing for the header inclusion.
+AddOutputFilter INCLUDES .html
+Options +Includes
+Options +FollowSymLinks
+
+# Make /foo.html available at /foo
+RewriteEngine on
+RewriteCond %{REQUEST_FILENAME}.html -f
+RewriteRule !.*\.html$ %{REQUEST_FILENAME}.html [L]
+
+# Turn on mod_pagespeed to optimize our docs.
+ModPagespeed on
+ModPagespeedRewriteLevel CoreFilters
+ModPagespeedInPlaceResourceOptimization on
--- a/doc/.htaccess
+++ b/doc/.htaccess
@@ -12,4 +12,5 @@ RewriteRule !.*\.html$ %{REQUEST_FILENAME}.html [L]
 ModPagespeed on
 ModPagespeedRewriteLevel CoreFilters
 ModPagespeedEnableFilters collapse_whitespace
+ModPagespeedEnableFilters remove_comments
 ModPagespeedInPlaceResourceOptimization on
--- a/doc/.htaccess
+++ b/doc/.htaccess
@@ -11,4 +11,5 @@ RewriteRule !.*\.html$ %{REQUEST_FILENAME}.html [L]
 # Turn on mod_pagespeed to optimize our docs.
 ModPagespeed on
 ModPagespeedRewriteLevel CoreFilters
+ModPagespeedEnableFilters collapse_whitespace
 ModPagespeedInPlaceResourceOptimization on
--- a/doc/_footer.html
+++ b/doc/_footer.html
@@ -39,8 +39,8 @@ function buildToc() {
   tocContents.textContent = "Contents";
   toc.appendChild(tocContents);
 
-  ulNodes = [];  // stack of active <ul> nodes
-  // replace level with ulNodes.length + 1
+  var level = 1;
+  var currentUl = null;
   for (var i = 0; i < headers.length; i++) {
     var header = headers[i];
     var headerLevel = header[0];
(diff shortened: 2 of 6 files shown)


doc/.htaccess
```diff
@@ -0,0 +1,14 @@
...
+AddOutputFilter INCLUDES .html
+Options +Includes
+Options +FollowSymLinks
+
+# Make /foo.html available at /foo
+RewriteEngine on
+RewriteCond %{REQUEST_FILENAME}.html -f
+RewriteRule !.*\.html$ %{REQUEST_FILENAME}.html [L]
+
+# Turn on mod_pagespeed to optimize our docs.
```
jmarantz, line 11: should we add collapse_whitespace?

jmarantz (commented):
What do you think about splitting this into two reviews.

In the first one, you pull over the django sources exactly, and I don't really review it :)

In the second one, you make the changes needed to render the documentation properly.

jeffkaufman:
@jmarantz 

>
>What do you think about splitting this into two reviews.
> * In the first one, you pull over the django sources exactly, and I don't really review it :)
> * In the second one, you make the changes needed to render the documentation properly.

Done.  I've split it into three reviews:

* #1464 : straight copy-over of existing docs
* #1465 : formatting changes that apply in bulk to all html files
* #1459 (this change) : css and js to make the docs be usable and nice to look at

doc/.htaccess
jeffkaufman: Sure, that's safe to add for our docs.

done: 8989e61

jeffkaufman:
@jlporter could you review the js in `_footer.html`?

doc/.htaccess
```diff
@@ -0,0 +1,15 @@
...
+ModPagespeed on
+ModPagespeedRewriteLevel CoreFilters
+ModPagespeedEnableFilters collapse_whitespace
```
jmarantz, line 14: add remove_comments.

doc/_footer.html
```diff
@@ -0,0 +1,72 @@
+<div id=footer>
+  <!--#include virtual="_navline.html" -->
+</div>
+
+<script>
+function build_toc_helper(node, headers) {
```
jmarantz: buildTocHelper

doc/_footer.html
```diff
@@ -0,0 +1,72 @@
...
+  if (node.nodeType == 1) {
+    // Element node.
+    var node_name = node.nodeName.toLowerCase();
+    if (node_name == "h2" || node_name == "h3" || node_name == "h4" ||
+        node_name == "h5" || node_name == "h6") {
+      if (node.id) {
+        headers.push([node_name, node.innerHTML, node.id]);
+        node.innerHTML +=
+          '&nbsp;<a class=header-link href="#' + node.id + '">&#x1f517;</a>';
```
jmarantz: Can we do this with DOM APIs rather than innerHTML?

doc/_footer.html
```diff
@@ -0,0 +1,72 @@
+<div id=footer>
+  <!--#include virtual="_navline.html" -->
+</div>
+
+<script>
+function build_toc_helper(node, headers) {
+  if (node.nodeType == 1) {
+    // Element node.
+    var node_name = node.nodeName.toLowerCase();
```
jmarantz: nodeName & other camel-case transforms here & below

doc/_footer.html
```diff
@@ -0,0 +1,72 @@
...
+  var level = 1;
+  for (var i = 0; i < headers.length; i++) {
+    var header = headers[i];
+    var header_level = header[0];
+    var header_value = header[1];
+    var header_id = header[2];
+
+    var new_level = parseInt(header_level.substring(1));
+    while (new_level > level) {
+      toc += "<ul>"
```
jmarantz: dom manipulation API

doc/.htaccess
jeffkaufman: done: 832bc0b39

doc/_footer.html
jeffkaufman: done: 26287b65b

doc/_footer.html
jeffkaufman: done: 26287b65b

doc/_footer.html
jeffkaufman: done: 79bb855a1

doc/_footer.html
jeffkaufman: done: 79bb855a1

jeffkaufman:
> At least add comments showing we are handling that case.

Done: 051489982

> What happens if you issue an alert if you detect a level gap? Does it fire?

I'd need to check each of our doc pages, which would be a pain; we have a lot of doc pages.

> Would that help keep the docs well-formed?

Probably not.  Either we'd use `alert()`, which would be a big pain for our users if we messed up, or we'd use `console.log()` and probably no one would notice.

jmarantz (approved):

PR #1465, opened 2017-01-03
doc: formatting html

jeffkaufman:
For all html files:

* Remove old meta paths
* Remove old `{% ... %}` and `{{ ...  }}`` directives from docs.
* Add `<h1>` titles
* Add an external stylesheet
* Wrap the contents in a div for styling
* Include headers and footers with ssi
* set a mobile viewport

This is a change to support #1459


doc/CVE-2012-4001.html
```diff
@@ -1,13 +1,15 @@
...
   <head>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>mod_pagespeed Security Advisory: Insufficient Hostname Verification</title>
-    <meta name="project_path" value="/speed/pagespeed/module/_project.yaml" />
-    <meta name="book_path" value="/speed/pagespeed/module/_book.yaml" />
-    <link rel="stylesheet" href="/speed/styles/local.css">
+    <link rel="stylesheet" href="doc.css">
   </head>
   <body>
+<!--#include virtual="_header.html" -->
```
jmarantz, line 8: ah....in this case I think that we should also enable remove_comments in .htacess, but that's in another pull request.

jmarantz (approved):

PR #1473, opened 2017-01-10
A simple script to collect slurp dumps for loadtest,

morlovich:
with help of PhantomJs

(Somewhat derived from the siege helper scripts)

### DIFF
--- a/devel/loadtest_collect/loadtest_collect.conf
+++ b/devel/loadtest_collect/loadtest_collect.conf
@@ -1,0 +1,42 @@
+# Global configuration.
+#
+# The ModPagespeedFileCachePath directory must exist and be writable
+# by the apache user (as specified by the User directive).
+ModPagespeedFileCachePath            "#HOME/apache2/pagespeed_cache/"
+
+# Direct Apache to send all HTML output to the mod_pagespeed
+# output handler.
+AddOutputFilterByType MOD_PAGESPEED_OUTPUT_FILTER text/html
+
+<VirtualHost localhost:8080>
+    # Turn on mod_pagespeed. To completely disable mod_pagespeed, you
+    # can set this to "off".
+    ModPagespeed on
+
+    <Location /pagespeed_admin>
+        Order allow,deny
+        Allow from localhost
+        Allow from 127.0.0.1
+        SetHandler pagespeed_admin
+    </Location>
+    <Location /pagespeed_global_admin>
+        Order allow,deny
+        Allow from localhost
+        Allow from 127.0.0.1
+        SetHandler pagespeed_global_admin
+    </Location>
+
+    KeepAlive On
+    KeepAliveTimeout 60
+
+    <Directory "#HOME/apache2/htdocs/" >
+        # This is enabled to make sure we don't crash mod_negotiation.
+        Options +MultiViews
+    </Directory>
+
+    ModPagespeedRewriteLevel AllFilters
+    ModPagespeedSlurpDirectory #SLURP_DIR
+    ModPagespeedSlurpReadOnly off
+    ModPagespeedRewriteDeadlinePerFlushMs -1
+    CustomLog "#LOG_PATH" %r
+</VirtualHost>
--- a/devel/loadtest_collect/loadtest_collect.conf
+++ b/devel/loadtest_collect/loadtest_collect.conf
@@ -1,16 +1,7 @@
-# Global configuration.
-#
-# The ModPagespeedFileCachePath directory must exist and be writable
-# by the apache user (as specified by the User directive).
 ModPagespeedFileCachePath            "#HOME/apache2/pagespeed_cache/"
-
-# Direct Apache to send all HTML output to the mod_pagespeed
-# output handler.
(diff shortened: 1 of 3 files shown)


jmarantz (commented):
Have you been able to run a load-test using the output of this?

jeffkaufman (commented):
Can you make this install phantomjs if we don't have it yet?

devel/loadtest_collect/loadtest_collect.conf
```diff
@@ -0,0 +1,42 @@
...
+# by the apache user (as specified by the User directive).
+ModPagespeedFileCachePath            "#HOME/apache2/pagespeed_cache/"
+
+# Direct Apache to send all HTML output to the mod_pagespeed
+# output handler.
+AddOutputFilterByType MOD_PAGESPEED_OUTPUT_FILTER text/html
+
+<VirtualHost localhost:8080>
+    # Turn on mod_pagespeed. To completely disable mod_pagespeed, you
+    # can set this to "off".
```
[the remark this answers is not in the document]
morlovich: Done

devel/loadtest_collect/loadtest_collect_corpus.sh
```diff
@@ -0,0 +1,61 @@
...
+# limitations under the License.
+#
+# This script collects slurps and URLs (post-optimization, if possible), of
+# some websites with the help of phantomjs.
+
+function usage {
+  echo "Usage: loadtest_collect/loadtest_collect_corpus.sh pages.txt out.tar.bz2"
+}
+
+set -u  # exit the script if any variable is uninitialized
```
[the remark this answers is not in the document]
morlovich: Seems brittle...

morlovich:
Having some difficulty commenting on other things directly, so let's just do it manually:
> Have you been able to run a load-test using the output of this?

Yep

> Can you make this install phantomjs if we don't have it yet?

For what platforms? Should it go through the fancy install script things?

jeffkaufman:
> For what platforms? Should it go through the fancy install script things?

Everything else with the dev environment assumes ubuntu 14, so it's fine with me if you just assume that as well.  At some point we may want to make the scripts under `devel/` support other distros, but for now I think people should just plan on using a VM.

morlovich:
phantomjs install, config comments cleanup pushed.

jeffkaufman (approved):
LGTM

devel/loadtest_collect/loadtest_collect_corpus.sh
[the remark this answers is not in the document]
jmarantz: Agreed, and not required for this CL.  I still think it'd be useful as a follow-up even if a "this is brittle" comment was put on it.

devel/loadtest_collect/loadtest_collect_corpus.sh
```diff
@@ -0,0 +1,65 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# This script collects slurps and URLs (post-optimization, if possible), of
+# some websites with the help of phantomjs.
+
+function usage {
+  echo "Usage: loadtest_collect/loadtest_collect_corpus.sh pages.txt out.tar.bz2"
```
jmarantz, line 21: add comment that pages.txt is a file of URLs, one per line?

devel/loadtest_collect/loadtest_collect_corpus.sh
```diff
@@ -0,0 +1,65 @@
...
+if [ -d devel ]; then
+  cd devel
+fi
+if [ ! -d loadtest_collect ]; then
+  echo Run this script from the top or devel/ directories
+  exit 1
+fi
+
+if [ ! $(which phantomjs) ]; then
+  sudo apt-get install phantomjs
```
jmarantz, line 43: about echoing what you are doing first.  If you run this from CentOS I think it'll just fail saying "apt-get command not found", and the user won't know it's just trying to install phantom without debugging the script.  With the echo at least the CentOS user could just install it manually.

jmarantz (approved):

devel/loadtest_collect/loadtest_collect_corpus.sh
morlovich: Done

devel/loadtest_collect/loadtest_collect_corpus.sh
morlovich: It's not quite that bad since you would get something like "sudo: apt-getttt: command not found"...
... but there is still a problem in that it just has sudo asking for a password for no good reason, so I've added an informational message. (or rather will have added once I push this, not at time of this comment)

(I am not a fan of the echo foo \n foo pattern since it duplicates the 'foo')

morlovich:
(PTAL at the tiny doc changes)

jeffkaufman:
> (PTAL at the tiny doc changes)

LGTM; ready to go in

PR #1474, opened 2017-01-11
build release tarball: open source it

jeffkaufman:
* change it from being given a branch to working on the current checkout
* change it to work with create_distro_tarball, which was already opensourced
* make it build openssl 1.0.2 if needed

### DIFF
--- a/devel/build_release_tarball.sh
+++ b/devel/build_release_tarball.sh
@@ -1,0 +1,93 @@
+#!/bin/bash
+#
+# Note: you might need to type your password a few times, once early, and once
+# at the end.
+#
+# This should be run on a release branch to make sure we can make a tarball and
+# at least build it on our workstations.  It will also copy the tarball into
+# ~/release (where the binaries usually go).
+#
+# Note that if this fails you may need to tweak the file list inside
+# devel/create_distro_tarball.sh
+
+set -e  # exit script if any command returns an error
+set -u  # exit the script if any variable is uninitialized
+
+function usage {
+  echo "Usage: devel/build_release_tarball.sh <beta|stable>"
+  exit 1
+}
+
+if [ $# -ne 1 ]; then
+  usage
+fi
+
+if [ ! -d net/instaweb ]; then
+  echo "This script must be run from the root of the mps checkout."
+  exit 1
+fi
+
+source net/instaweb/public/VERSION
+RELEASE="$MAJOR.$MINOR.$BUILD.$PATCH"
+CHANNEL="$1"
+
+deps="libpng12-dev libicu-dev libssl-dev libjpeg-dev realpath build-essential
+      pkg-config gperf unzip libapr1-dev libaprutil1-dev apache2-dev"
+if dpkg-query -Wf '${Status}\n' $deps 2>&1 | \
+     grep -v "install ok installed"; then
+  # Only run apt-get install if one of the deps is not already installed.
+  # See: http://stackoverflow.com/questions/1298066
+  sudo apt-get install $deps
+fi
+
+RELEASE_DIR="$HOME/release/$RELEASE"
+mkdir -p "$RELEASE_DIR"
+REVISION="$(build/lastchange.sh "$PWD" | sed 's/LASTCHANGE=//')"
+TARBALL="$RELEASE_DIR/mod-pagespeed-$CHANNEL-$RELEASE-r$REVISION.tar.bz2"
+devel/create_distro_tarball.sh "$TARBALL"
+
+echo "Tarball should now be at $TARBALL"
+
+# Try to build it
+BUILD_DIR="$(mktemp -d)"
+echo "Doing a test build inside $BUILD_DIR"
+cd "$BUILD_DIR"
+
+if openssl version | grep "^OpenSSL 1[.]0[.][01]\|^OpenSSL 0[.]"; then
+  echo "Your openssl version is too old to build the tarball; we need 1.0.2+"
+  echo "Building 1.0.2 from source..."
+  OPENSSL_VERSION="1.0.2j"
+  wget "https://www.openssl.org/source/openssl-${OPENSSL_VERSION}.tar.gz"
+  tar -xzvf "openssl-${OPENSSL_VERSION}.tar.gz"
+  cd openssl-"${OPENSSL_VERSION}"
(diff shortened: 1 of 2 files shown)


jeffkaufman:
Once this is in I'll backport it to branch 34

devel/build_release_tarball.sh
```diff
@@ -0,0 +1,94 @@
...
+deps="libpng12-dev libicu-dev libssl-dev libjpeg-dev realpath build-essential
+      pkg-config gperf unzip libapr1-dev libaprutil1-dev apache2-dev"
+if dpkg-query -Wf '${Status}\n' $deps 2>&1 | \
+     grep -v "install ok installed"; then
+  # Only run apt-get install if one of the deps is not already installed.
+  # See: http://stackoverflow.com/questions/1298066
+  sudo apt-get install $deps
+fi
+
+# Checkout everything, then tarball the parts that we want.
```
morlovich: Does it actually check anything out here?

devel/build_release_tarball.sh
```diff
@@ -0,0 +1,94 @@
...
+devel/create_distro_tarball.sh "$TARBALL"
+
+echo "Tarball should now be at $TARBALL"
+
+# Try to build it
+BUILD_DIR="$(mktemp -d)"
+echo "Doing a test build inside $BUILD_DIR"
+cd "$BUILD_DIR"
+
+if openssl version | grep "^OpenSSL 1[.]0[.][01]"; then
```
morlovich: What with this do with OpenSSL 0.9.x, etc, or 1.1.0?

devel/build_release_tarball.sh
```diff
@@ -0,0 +1,94 @@
...
+cd modpagespeed*
+./generate.sh -Dsystem_include_path_apr=/usr/include/apr-1.0/ \
+              -Dsystem_include_path_httpd=/usr/include/apache2
+cd src
+make -j6
+out/Debug/mod_pagespeed_test
+# These tests fail because they are golded against a specific version of
+# compression libraries.
+# TODO(sligocki): Could we change the tests to be less fragile or test in a
+# different way in this case?
```
morlovich, line 85: I did do some of that in Debian package patches, though not sure exactly for these ones.

devel/create_distro_tarball.sh
```diff
@@ -65,163 +66,130 @@ src/build/gyp_chromium -D use_system_libs=1 \$*
...
-    cd $CHECKOUT_DIR
-    git clone https://github.com/pagespeed/mod_pagespeed.git src
-    cd src
-    git checkout "$GIT_TAG"
-    git submodule update --init --recursive
-  fi
-else
-  SRC_DIR=$2
-fi
+git submodule update --init --recursive
```
morlovich, line 88: Hmm, why keep this?
It does make sense to get rid of checkout stuff here at all, what with git now being the dev environment...

devel/create_distro_tarball.sh
```diff
@@ -65,163 +66,130 @@ src/build/gyp_chromium -D use_system_libs=1 \$*
...
+    $DIR/third_party/serf \
+    $DIR/third_party/zlib/zlib.gyp \
+    $DIR/tools/gyp \
+    $DIR/url
 
-if [ x != x$CHECKOUT_DIR ]; then
-  echo "You may want to rm -rf $CHECKOUT_DIR to clean up"
-fi
+cd "$MPS_CHECKOUT"
+rm -r "$TEMP_DIR"
```
morlovich, line 195: Did you compare contents of tarballs?

devel/create_distro_tarball.sh
jeffkaufman: We need this because otherwise if someone runs this in a clean client it will fail to find all the deps.

(I've tested this by getting a raw Ubuntu 14 LTS VM and running "git clone mps" and then running this script, which now succeeds.  In addition to verifying that it succeeds on our workstations.)

devel/build_release_tarball.sh
jeffkaufman: Ack

devel/build_release_tarball.sh
jeffkaufman: Fixed to also object to 0.x.

1.1.0 is fine for us, and this grep won't object to it.

devel/build_release_tarball.sh
jeffkaufman: sorry, that comment isn't needed anymore; fixed

devel/create_distro_tarball.sh
jeffkaufman: Yes, it looks reasonable.  The main change is that `.mk` and `.pyc` files aren't included anymore, which I think was one of your changes?

jeffkaufman:
@morlovich ready for another round of review

morlovich:
I don't think you actually pushed it?

jeffkaufman:
@morlovich 
> I don't think you actually pushed it?

Sorry, I did push it, but I forced pushed it because I wanted to squash down to one commit to cherry-pick onto 34 for testing.

morlovich:
I think you could have squashed it afterwards, at the merge step? Anyway, Looking..

devel/build_release_tarball.sh
```diff
@@ -0,0 +1,93 @@
...
+#
+# Note: you might need to type your password a few times, once early, and once
+# at the end.
+#
+# This should be run on a release branch to make sure we can make a tarball and
+# at least build it on our workstations.  It will also copy the tarball into
+# ~/release (where the binaries usually go).
+#
+# Note that if this fails you may need to tweak the file list inside
+# devel/create_distro_tarball.sh
```
morlovich, line 14: It may be worth documenting what distro this assumes.

morlovich (approved):

devel/build_release_tarball.sh
jeffkaufman: You mean that this only runs on Ubuntu 14 right now, like the rest of our dev environment?

morlovich: Yeah, mostly due to the package install step.

jeffkaufman: done: 2f0d78af7

jeffkaufman:
Backported to branch 34 as 64885264800924420eef235b366b2ab8f8e046cd

PR #1476, opened 2017-01-12
doc: use HTTPS for LazyloadImagesBlankUrl

jeffkaufman:
Fixes https://github.com/pagespeed/ngx_pagespeed/issues/1358


morlovich:
Maybe protocol relative?

jeffkaufman:
> Maybe protocol relative?

The recommendation now is just to always use https: https://www.paulirish.com/2010/the-protocol-relative-url/

```
Update 2014.12.17:

Now that SSL is encouraged for everyone and
doesn’t have performance concerns, this technique
is now an anti-pattern. If the asset you need is
available on SSL, then always use the https:// asset.
```

jmarantz (approved):

PR #1482, opened 2017-01-19
standby: add standby mode

jeffkaufman:
In standby mode PageSpeed is off, except it serves .pagespeed. resources and
PageSpeed query parameters are interpreted.  This is equivalent to "off" in
mod_pagespeed.

With this change, "off" in mod_pagespeed is deprecated, and people should use
"standby" instead.

See also: https://github.com/pagespeed/ngx_pagespeed/pull/1365

Fixes: https://github.com/pagespeed/ngx_pagespeed/issues/968

### DIFF
--- a/html/doc/configuration.html
+++ b/html/doc/configuration.html
@@ -97,8 +97,15 @@ ModPagespeed off
 <pre class="prettyprint">
 ModPagespeed unplugged
 </pre>
+
 <p>
-The <code>on</code> and <code>off</code> values can be used in
+
+In versions UNRELEASED and later "off" is deprecated and "standby" has been
+added as a replacement.
+
+<p>
+The <code>on</code>, <code>off</code>, and (in UNRELEASED and
+later) <code>standby</code> values can be used in
 <a href="#htaccess"><code>.htaccess</code> files,
 <code>&lt;Directory&gt;</code></a> scopes, <code>query parameters</code>, and
 <code>headers</code>.  The <code>unplugged</code> value can only be used in the
@@ -166,6 +173,15 @@ pagespeed off;
 </pre>
 This is equivalent to the "unplugged" setting for mod_pagespeed, and entirely
 disables ngx_pagespeed.
+<p>
+
+As of UNRELEASED, "off" is deprecated, and replaced with the synonym
+"unplugged".  Additionally the "standby" setting has been added, which is
+equivalent to the "off" or "standby" setting for mod_pagespeed.
+
+<p class=warning><b>Warning:</b> do not set ngx_pagespeed to "unplugged" mode in
+1.12.34.1 and earlier.  That will trigger a crash on every request.  With those
+versions, use "off" instead.
 
 <h2 id="respectvary">Respecting Vary Headers</h2>
 <p>
--- a/html/doc/configuration.html
+++ b/html/doc/configuration.html
@@ -59,6 +59,86 @@ See the <a href="admin#config">Admin Page documentation</a> for
 instructions on how to configure handlers to provide visibility and
 control into PageSpeed's operation.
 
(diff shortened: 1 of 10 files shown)


html/doc/configuration.html
```diff
@@ -166,6 +173,15 @@ <h3 id="on_off_nginx">Turning the module on and off</h3>
 </pre>
 This is equivalent to the "unplugged" setting for mod_pagespeed, and entirely
 disables ngx_pagespeed.
+<p>
+
+As of UNRELEASED, "off" is deprecated, and replaced with the synonym
+"unplugged".  Additionally the "standby" setting has been added, which is
+equivalent to the "off" or "standby" setting for mod_pagespeed.
```
morlovich: I am having a hard time following this. A table may be in order...

jeffkaufman: Ok, I've rewritten the doc pretty thoroughly.  What do you think now?

morlovich:
That does look great.

jeffkaufman:
LGTY?

jeffkaufman:
@morlovich ping

html/doc/configuration.html
```diff
@@ -107,6 +139,22 @@ <h3 id="on_off">Turning the module on and off</h3>
 unplugged directive.
```
morlovich: unplugged directive?

net/instaweb/rewriter/rewrite_query.cc
```diff
@@ -545,7 +545,7 @@ RewriteQuery::Status RewriteQuery::ScanNameValue(
       if (pairs[i] == "no-transform") {
         // TODO(jmarantz): A .pagespeed resource should return un-optimized
         // content with "Cache-Control: no-transform".
-        options->set_enabled(RewriteOptions::kEnabledOff);
+        options->set_enabled(RewriteOptions::kEnabledStandby);
```
morlovich, line 548: Hmm, the story with query param = off (or standby in crashy vers) needs double-checking.

jeffkaufman: What are you concerned about?  If we get as far as interpreting query parameters then we're definitely not unplugged, so standby seems right to me,.

html/doc/configuration.html
jeffkaufman: fixed: 0d4c70d2e

jeffkaufman:
Passes checkin.

net/instaweb/rewriter/rewrite_query.cc
morlovich: I meant what happens if "off" or "unplugged" are query params... what I forgot is that IIRC we don't do that for .pagespeed. resources anyway?

jeffkaufman: We do accept `?PageSpeed=off` on resources; that's one way to see what a resource would look like without IPRO.

But I'm still not clear on what you're objecting to here.  Would it help to talk in person?

morlovich:
Not really objecting, rather asking for clarification of what "=off" does
on query params.

jeffkaufman:
> Not really objecting, rather asking for clarification of what `"=off"` does on query params.

If you set `PageSpeed=off` as a query param, we create custom options with PageSpeed disabled and don't optimize the response at all.  This PR doesn't change that either way.

morlovich:
LGTM modulo preferring more tests on query param = off

jeffkaufman:
> LGTM modulo preferring more tests on query param = off

Are you counting all the existing tests that use `?PageSpeed=off` and would stop passing if it stopped working?  Like `automatic/system_tests/initial_header_check.s`?

jeffkaufman:
@morlovich confirmed offline that he's ready for this to go in

PR #1492, opened 2017-02-09
IPRO+LFF: content length & handling of unoptimizable resources (draft)

oschaaf:
Putting this up here in a non-commitable state for early consideration/review.

Contains fixes in the IPRO flow for resources covered by LoadFromFile:

1. Attempts to fix s-maxage getting stuck: https://github.com/pagespeed/mod_pagespeed/issues/1485  
As a bonus this saves repeated cache writes to our own cache when this flow is hit.

2. Avoid chunked transfer-encodings in this flow by writing out a content-length header : https://github.com/pagespeed/mod_pagespeed/issues/1240

Not ready for committing yet: 
- It would be good to add a couple of tests to avoid regressions
- This needs to be double checked if and how this interacts with the (gzip) compression feature 
- Need to check if adding a Vary: header is appropriate in this flow, at this stage, while I'm at it. (because afaict it is missing in this flow).

Note: passes the mps smoke tests and the system tests on my machine

### DIFF
--- a/net/instaweb/rewriter/in_place_rewrite_context.cc
+++ b/net/instaweb/rewriter/in_place_rewrite_context.cc
@@ -103,6 +103,7 @@ RecordingFetch::~RecordingFetch() {}
 void RecordingFetch::HandleHeadersComplete() {
   can_in_place_rewrite_ = CanInPlaceRewrite();
   streaming_ = ShouldStream();
+
   if (can_in_place_rewrite_) {
     // Save the headers, and wait to finalize them in HandleDone().
     saved_headers_.reset(new ResponseHeaders(*response_headers()));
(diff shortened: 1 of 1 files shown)


oschaaf:
This PR attempts to avoid recording in the IPRO path for every request in the case where:
- The url is mapped by LoadFromFile
- Earlier attempts for the same url have been marked as unoptimizeable.

Does anyone have any thoughts on wether this PR addresses this in the right place/way?:

Instead of using `RecordingFetch`, the result of `resource->LoadAsync(...)` is passed on directly in these cases. 

Alternatively, I've dumped state [1] on a simple PoC for a `LoadFromFileFetcher`. That would avoid having to make distinctions like this altogether in a lot of places (and have the LoadFromFile inputs leverage our caching). We could maybe use inotify to watch for changes in LoadFromFile-covered paths to purge the cache when appropriate to preserve the instant-update perk that currently comes with LoadFromFile.

[1] https://github.com/pagespeed/mod_pagespeed/commit/19213d5412d477313e10bf29a9ffc1faf4b576aa

PR #1494, opened 2017-02-14
CSP: Parse details in host-scheme

morlovich:
This is a bit complicated since those aren't URLs proper, as they can have wildcards in spot and 
can drop the scheme.

### DIFF
--- a/net/instaweb/rewriter/csp.cc
+++ b/net/instaweb/rewriter/csp.cc
@@ -49,43 +49,139 @@ inline bool IsAsciiAlpha(char ch) {
           ((ch >= 'A') && (ch <= 'Z')));
 }
 
+inline bool IsAsciiDigit(char ch) {
+  return ((ch >= '0') && (ch <= '9'));
+}
+
 }  // namespace
+
+bool CspSourceExpression::TryParseScheme(StringPiece* input) {
+  if (input->size() < 2) {
+    // Need at least a: or such.
+    return false;
+  }
+  
+  if (!IsAsciiAlpha((*input)[0])) {
+    return false;
+  }
+  
+  size_t pos = 1;
+  while (pos < input->size() &&
+         (IsAsciiAlphaNumeric((*input)[pos]) 
+          || (*input)[pos] == '+' 
+          || (*input)[pos] == '-' 
+          || (*input)[pos] == '.')) {
+    ++pos;
+  }
+  
+  if (pos == input->size() || (*input)[pos] != ':') {
+    // All schema characters, but no : or :// afterwards -> something else.
+    return false;
+  }
+  
+  if (pos == (input->size() - 1)) {
+    // Last character was also : -> clearly a scheme-source
+    kind_ = kSchemeSource;
+    input->substr(0, pos).CopyToString(&mutable_url_data()->scheme_part);
+    input->remove_prefix(pos + 1);
+    return true;
+  }
+  
+  // We now want to see if it's actually foo://
+  if ((pos + 2 < input->size())
+       && ((*input)[pos + 1] == '/') && ((*input)[pos + 2] == '/')) {
+    input->substr(0, pos).CopyToString(&mutable_url_data()->scheme_part);
+    input->remove_prefix(pos + 3);
+  }
+  
+  // Either way, it's not a valid scheme-source at this point, even if it's
+  // a valid host-source
+  return false;
+}
 
 CspSourceExpression CspSourceExpression::Parse(StringPiece input) {
   TrimCspWhitespace(&input);
   if (input.empty()) {
-    return CspSourceExpression(kUnknown);
+    return CspSourceExpression();
   }
 
   if (input.size() > 2 && input[0] == '\'' && Last(input) == '\'') {
     return ParseQuoted(input.substr(1, input.size() - 2));
   }
-
-  // Check for scheme-source.
-  if (input.size() >= 2 && Last(input) == ':') {
-    // scheme = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )
-    bool is_scheme = true;
-    if (IsAsciiAlpha(input[0])) {
-      for (size_t i = 1; i < (input.size() - 1); ++i) {
-        char c = input[i];
-        if (!IsAsciiAlphaNumeric(c) && (c != '+') && (c != '-') && (c != '.')) {
-          is_scheme = false;
-          break;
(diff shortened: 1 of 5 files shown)


net/instaweb/rewriter/csp.cc
```diff
@@ -49,43 +49,139 @@ inline bool IsAsciiAlpha(char ch) {
...
+  // Either way, it's not a valid scheme-source at this point, even if it's
+  // a valid host-source
+  return false;
+}
+
 CspSourceExpression CspSourceExpression::Parse(StringPiece input) {
   TrimCspWhitespace(&input);
   if (input.empty()) {
-    return CspSourceExpression(kUnknown);
+    return CspSourceExpression();
```
oschaaf, line 101: I wonder, will we need to be able to discern between unknown and empty/whitespace ?

net/instaweb/rewriter/csp_test.cc
```diff
@@ -65,31 +85,119 @@ TEST(CspParseSourceTest, NonQuoted) {
...
+      CspSourceExpression(
+          CspSourceExpression::kHostSource,
+          CspSourceExpression::UrlData("https", "*", "443", "/foo.js")),
+      CspSourceExpression::Parse("https://*:443/foo.js"));
+
+  EXPECT_EQ(
+      CspSourceExpression(CspSourceExpression::kUnknown),
+      CspSourceExpression::Parse("https://*:443?foo.js"));
+}
+
```
oschaaf, line 196: Maybe also test our expectations for `CspSourceExpression::ParseQuoted("\"\"")` ?

net/instaweb/rewriter/csp.cc
morlovich: Nope, basically either one won't authorize anything.

net/instaweb/rewriter/csp_test.cc
morlovich: Double quotes don't mean anything in this syntax, so any particular reason this would be interesting?

net/instaweb/rewriter/csp.cc
oschaaf: I see, thanks

net/instaweb/rewriter/csp_test.cc
oschaaf: `CspSourceExpression::ParseQuoted("\"\"")` -> should have been `CspSourceExpression::ParseQuoted("''")`
But either way, I noticed there's already a test that inputs only whitespace, and this one wouldn't add value on top of that, so please ignore

morlovich: Oh, good idea, this is an edge case, so added.

morlovich:
Note: there is a more extensive/later version in https://github.com/pagespeed/mod_pagespeed/tree/morlovich-csp-urls3, it might make sense to review that instead.

net/instaweb/rewriter/csp.cc
```diff
@@ -49,43 +49,139 @@ inline bool IsAsciiAlpha(char ch) {
           ((ch >= 'A') && (ch <= 'Z')));
 }
 
+inline bool IsAsciiDigit(char ch) {
+  return ((ch >= '0') && (ch <= '9'));
+}
```
jmarantz: put this in string_util.h next to IsHexDigit?  Rename to IsDecimalDigit?

jmarantz (commented):
qq: have you valgrind'd & msan'd your tests for this?  I don't see anything suspicious but it seems worthwhile.

also maybe fuzzing?

still reading...

morlovich:
Hmm, I've done fuzzing, but probably not in the latest revision ---
basically by copying the files to the Google build system and using the
support there.
Can also copy them to the Chrome build system and do the same :). Probably
should redo for latest.

net/instaweb/rewriter/csp.cc
```diff
@@ -49,43 +49,139 @@ inline bool IsAsciiAlpha(char ch) {
...
+  if (!IsAsciiAlpha((*input)[0])) {
+    return false;
+  }
+
+  size_t pos = 1;
+  while (pos < input->size() &&
+         (IsAsciiAlphaNumeric((*input)[pos])
+          || (*input)[pos] == '+'
+          || (*input)[pos] == '-'
+          || (*input)[pos] == '.')) {
```
jmarantz: it annoys me slightly that (*input)[pos] is repeated 4x :)

Maybe just factor out the per-char check to an inline helper?

net/instaweb/rewriter/csp.cc
```diff
@@ -49,43 +49,139 @@ inline bool IsAsciiAlpha(char ch) {
...
+    }
+
+    if (host_part.starts_with("*")) {
+      if (!host_part.starts_with("*.") && (host_part != "*")) {
+        return CspSourceExpression();
+      }
+    }
+
+    // Start on port-part, if any
+    if (input.starts_with(":")) {
```
jmarantz: input[0] == ':'

net/instaweb/rewriter/csp.cc
```diff
@@ -49,43 +49,139 @@ inline bool IsAsciiAlpha(char ch) {
...
+    }
+
+    // Verify accumulated host-part is valid.
+    StringPiece host_part(result.url_data().host_part);
+    if (host_part.empty()) {
+      return CspSourceExpression();
+    }
+
+    if (host_part.starts_with("*")) {
+      if (!host_part.starts_with("*.") && (host_part != "*")) {
```
jmarantz: if (host_part[0] == '*' && host_part.size() > 1 && host_part[1] != '.' )

net/instaweb/rewriter/csp.cc
morlovich: input could be empty here.

jmarantz (approved):
It'd be great to get some coverage tests on this for your unit tests.

net/instaweb/rewriter/csp.cc
morlovich: Done

net/instaweb/rewriter/csp.cc
morlovich: Donr

net/instaweb/rewriter/csp.cc
morlovich: Thanks, this is much better.

PR #1495, opened 2017-02-16
Add an anchor for "Configuring SSL Certificates"

oschaaf:


morlovich (approved):
Thanks! I think the JS will add the nice little chain icon, too.

oschaaf:
Yes the chain icon is rendered.

PR #1497, opened 2017-02-20
Clean the response headers when sending kNotInCache

oschaaf:
When we send out a 501 in the IPRO path for a http cache
hit, clear the response headers so we do not leave
behind stray headers.

Fixes https://github.com/pagespeed/mod_pagespeed/issues/1496

### DIFF
--- a/net/instaweb/rewriter/in_place_rewrite_context.cc
+++ b/net/instaweb/rewriter/in_place_rewrite_context.cc
@@ -139,6 +139,7 @@ void RecordingFetch::HandleHeadersComplete() {
       // kNotInCacheStatus instead to fall back to the server's native method of
       // serving the url and indicate we do want it recorded.
       if (!response_headers()->IsErrorStatus()) {
+        response_headers()->Clear();
         response_headers()->set_status_code(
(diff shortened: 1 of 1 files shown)


morlovich:
So this seems to work differently in nginx, since in MPS the 501 will never see the wire; is the difference because you're actually doing the initial check asynchronously?

oschaaf:
In ngx_pagespeed the the headers from the IPRO lookup are (unconditionally) copied to the request header structure associated to nginx's request. Arguably that structure should be cleaned before continuing with processing the response in case the response code is a 501. 
We could do that here:
https://github.com/pagespeed/ngx_pagespeed/blob/master/src/ngx_pagespeed.cc#L2500
Maybe we shouldn't even copy these headers at all when we are not planning to put them on the wire and set a flag or status instead.

On the other hand, I thought would be nice if PSOL writes out a clean 501 response header, as that makes it impossible to run into this, and it is also seemed the simpler fix..

morlovich:
I think this is fine, but it's probably worth adding a comment near the clear call explaining why it's needed.

oschaaf:
Done

morlovich:
LGTM, Thanks

PR #1499, opened 2017-02-21
Disable CSS (really font CSS in practice) inlining on AMP.

morlovich:
Ref: https://github.com/pagespeed/ngx_pagespeed/issues/1382

### DIFF
--- a/net/instaweb/rewriter/css_inline_filter.cc
+++ b/net/instaweb/rewriter/css_inline_filter.cc
@@ -117,6 +117,16 @@ void CssInlineFilter::EndElementImpl(HtmlElement* element) {
   const char* media = NULL;
   if (CssTagScanner::ParseCssElement(element, &href, &media) &&
       !driver()->HasChildrenInFlushWindow(element)) {
+    if (driver()->is_amp_document()) {
+      // Don't inline into AMP documents. Those do permit font loading CSS,
+      // which we could in principle inline, but they also restrict the document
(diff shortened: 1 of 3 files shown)


oschaaf:
fwiw, looks good to me

morlovich:
Well, that failure sure looks relevant, trying to see if I can reproduce it locally...

oschaaf:
@morlovich ... It sure does, yes.

But did you note the other failure from `continuous-integration/travis-ci/push`? 
```
(98)Address already in use: make_sock: could not bind to address 127.0.0.1:8087
```
I wonder if something odd happened with these test runs..
Do you know if there is a way to respin these tests from the start (aside from pushing a change)?

morlovich:
I think there may be a way in the Travis UI, but I don't feel comfortable giving it the permissions it wants...

morlovich:
Well, it passed devel/checkin, FWIW...

morlovich:
Hmm, saw the same failure as a flake on a totally different CL. I think I'll re-read the test...

morlovich:
Huh, they're different a bit. In this one we seem to have had the font CSS inlined in fetch_util but not in the next line over wget_dump. The other one seem to have run into an experiment which tweaked the output some. I think I'll change the tests to at least use the same output for consistency, and change the comparison op to >= rather than =

morlovich:
OK, maybe this will be a bit less flakey, PTAL.

jmarantz (approved):

PR #1502, opened 2017-02-27
CSP prep: add a notion of input role, and wire it to CreateInputResource

morlovich:
This is needed to eventually know which policy to check.
(The enum is similar to SemanticRole, but it was made separate to give
 the cases other than image/style/script clear semantics).

This change itself should be a no-op functionally.

OutputResource methods will likely need a similar treatment.

### DIFF
--- a/net/instaweb/rewriter/render_blocking_html_computation.cc
+++ b/net/instaweb/rewriter/render_blocking_html_computation.cc
@@ -100,8 +100,11 @@ void RenderBlockingHtmlComputation::Compute(const GoogleString& url) {
   }
 
   bool is_authorized = false;
+  // Note: this is for HTML, so may be kReconstruction would be appropriate,
+  // but it's tough to audit since this doesn't seem to have any users?
   ResourcePtr resource(
-      parent_driver_->CreateInputResource(gurl, &is_authorized));
+      parent_driver_->CreateInputResource(
+          gurl, RewriteDriver::InputRole::kUnknown, &is_authorized));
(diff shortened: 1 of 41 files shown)


oschaaf:
lgtm, no comments

net/instaweb/rewriter/render_blocking_html_computation.cc
```diff
@@ -100,8 +100,11 @@ void RenderBlockingHtmlComputation::Compute(const GoogleString& url) {
   }
 
   bool is_authorized = false;
+  // Note: this is for HTML, so may be kReconstruction would be appropriate,
+  // but it's tough to audit since this doesn't seem to have any users?
```
jmarantz, line 104: This is suspected dead code?  What happens if you put CHECK(false) and run tests?

jmarantz (approved):

net/instaweb/rewriter/render_blocking_html_computation.cc
morlovich: Did something different: If I comment out render_blocking_html_computation.cc and render_blocking_html_computation_test.cc in their appropriate .gyp files the whole thing compiles & links & runs unit tests just fine.

I think this was probably used by the mobilizer, shall I remove it?

jmarantz: Yes, remove it.  I'm not sure what it was used for but it's time to bring out our dead :)

PR #1506, opened 2017-03-04
doc: remove non-ascii characters

jeffkaufman:
The documentation html has a small number of non-ascii characters that aren't displayed properly on modpagespeed.com.  For example, see https://modpagespeed.com/doc/system#ipro right before 'cases where'. While serving the page unicode-compatible would be a better fix, we don't actually need any of these.  This PR just switches them for the ascii versions.


jeffkaufman:
Signed the CLA

jmarantz (approved):
thanks Jeff!

PR #1510, opened 2017-03-09
Expose try_best_compression_for_png option

huibaolin:


oschaaf:
LGTM. Out of curiosity, is the intent to wire this setting completely through so that it can be configured?

huibaolin:
Yes, this is to enable some products to optimize PNG faster but with less-than-optimal compression ratio.

PR #1512, opened 2017-03-13
CSP:  Implements the algorithm for matching URLs against source patterns.

morlovich:
This also includes doing some canonicalization and representation tweaks on the parsed version to make comparisons easier.

### DIFF
--- a/net/instaweb/rewriter/csp.cc
+++ b/net/instaweb/rewriter/csp.cc
@@ -199,8 +199,8 @@ bool CspSourceExpression::Matches(
   // Implementation of the "Does url match expression in origin with
   // redirect count?" algorithm (where redirect count is 0 for our
   // purposes, since we check the request).
-  // TODO(morlovich): This means redirect following may require changes
-  // here.
+  // https://w3c.github.io/webappsec-csp/#match-url-to-source-list
+
   if (kind_ != kSelf && kind_ != kSchemeSource && kind_ != kHostSource) {
     return false;
   }
@@ -304,7 +304,9 @@ bool CspSourceExpression::Matches(
     }
   }
 
-  if (!expr_path.empty()) {  // this would also be skipped for redirects
+  // TODO(morlovich):Redirect following may require changes here ---
+  // this would also be skipped for redirects.
+  if (!expr_path.empty()) {
     // TODO(morlovich): Verify that behavior for query here is what we want.
     StringPieceVector url_path_list;
     SplitStringPieceToVector(url.PathAndLeaf(), "/", &url_path_list, true);
@@ -422,7 +424,8 @@ std::unique_ptr<CspPolicy> CspPolicy::Parse(StringPiece input) {
       if (dir_name != CspDirective::kNumSourceListDirectives &&
           policy->policies_[static_cast<int>(dir_name)] == nullptr) {
         // Note: repeated directives are ignored per the "Parse a serialized
-        // CSP as disposition" algorith,
+        // CSP as disposition" algorithm.
+        // https://w3c.github.io/webappsec-csp/#parse-serialized-policy
         policy->policies_[static_cast<int>(dir_name)]
             = CspSourceList::Parse(value);
       }
--- a/net/instaweb/rewriter/csp.cc
+++ b/net/instaweb/rewriter/csp.cc
@@ -80,6 +80,7 @@ bool CspSourceExpression::TryParseScheme(StringPiece* input) {
     kind_ = kSchemeSource;
     input->substr(0, pos).CopyToString(&mutable_url_data()->scheme_part);
     input->remove_prefix(pos + 1);
+    LowerString(&mutable_url_data()->scheme_part);
     return true;
   }
 
@@ -88,6 +89,7 @@ bool CspSourceExpression::TryParseScheme(StringPiece* input) {
        && ((*input)[pos + 1] == '/') && ((*input)[pos + 2] == '/')) {
     input->substr(0, pos).CopyToString(&mutable_url_data()->scheme_part);
     input->remove_prefix(pos + 3);
+    LowerString(&mutable_url_data()->scheme_part);
   }
 
   // Either way, it's not a valid scheme-source at this point, even if it's
@@ -136,6 +138,7 @@ CspSourceExpression CspSourceExpression::Parse(StringPiece input) {
       result.mutable_url_data()->host_part.push_back(input[0]);
       input.remove_prefix(1);
     }
+    LowerString(&result.mutable_url_data()->host_part);
 
     // Verify accumulated host-part is valid.
     StringPiece host_part(result.url_data().host_part);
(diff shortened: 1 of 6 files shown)


net/instaweb/rewriter/csp.cc
```diff
@@ -172,12 +175,158 @@ CspSourceExpression CspSourceExpression::Parse(StringPiece input) {
...
+  // Implementation of the "Does url match expression in origin with
+  // redirect count?" algorithm (where redirect count is 0 for our
+  // purposes, since we check the request).
+  // TODO(morlovich): This means redirect following may require changes
+  // here.
+  if (kind_ != kSelf && kind_ != kSchemeSource && kind_ != kHostSource) {
+    return false;
+  }
+
+  if (!origin_url.IsAnyValid() || !url.IsAnyValid()) {
```
oschaaf, line 208: Do we need to specifically handle blob / data / filesystem schemes here?

net/instaweb/rewriter/csp_test.cc
```diff
@@ -186,12 +194,214 @@ TEST(CspParseSourceTest, NonQuoted) {
...
+  CheckMatch(true, "*", "gopher://origin", "ftp://www.example.com");
+
+  // Oddly, as spec'd, this doesn't include ws: and wss:
+  CheckMatch(false, "*", "gopher://origin", "ws://www.example.com");
+  CheckMatch(false, "*", "gopher://origin", "wss://www.example.com");
+
+  // Other schemes have to match origin to be permitted.
+  CheckMatch(true, "*", "gopher://origin", "gopher://www.example.com");
+  CheckMatch(false, "*", "gopher://origin", "weirder://www.example.com");
+}
```
oschaaf, line 262: Maybe also test assumptions on data / blob / filesystem uri's?

net/instaweb/rewriter/csp.cc
```diff
@@ -172,12 +175,158 @@ CspSourceExpression CspSourceExpression::Parse(StringPiece input) {
...
+    // TODO(morlovich): Check whether the :80/:443 case is about effective
+    // or explicit port.
+    if (expr_port != "*"
+        && expr_port != IntegerToString(url.EffectiveIntPort())
+        && !(expr_port == "80" && url.EffectiveIntPort() == 443)) {
+      return false;
+    }
+  }
+
+  if (!expr_path.empty()) {  // this would also be skipped for redirects
```
morlovich: This is the only spot that will need to be changed for redirects --- basically steps after redirect don't do the path check.

net/instaweb/rewriter/public/csp.h
```diff
@@ -42,29 +49,41 @@ class CspSourceExpression {
...
+                    " path_exact_match:", BoolToString(path_exact_match));
     }
 
     bool operator==(const UrlData& other) const {
       return scheme_part == other.scheme_part &&
              host_part == other.host_part &&
              port_part == other.port_part &&
-             path_part == other.path_part;
+             path_part == other.path_part &&
+             path_exact_match == other.path_exact_match;
```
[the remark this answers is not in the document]
morlovich: This is just for unit tests.

net/instaweb/rewriter/csp.cc
oschaaf: re: "This is the only spot that will need to be changed for redirects --- basically steps after redirect don't do the path check." 
Ok, thanks for clearing that up

net/instaweb/rewriter/csp.cc
```diff
@@ -172,12 +175,158 @@ CspSourceExpression CspSourceExpression::Parse(StringPiece input) {
...
+        !input.empty() && !input.ends_with("/");
   }
 
   return result;
 }
 
+bool CspSourceExpression::Matches(
+    const GoogleUrl& origin_url, const GoogleUrl& url) const {
+  // Implementation of the "Does url match expression in origin with
+  // redirect count?" algorithm (where redirect count is 0 for our
```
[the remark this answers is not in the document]
morlovich: Done

net/instaweb/rewriter/csp.cc
morlovich: Hmm. Actually I want to limit this to http[s], since I dropped support for ws/wss (as we don't use them and they were making for some complicated expressions)

morlovich: Actually probably want data:, too, assuming we actually fetch it (checking)

morlovich: We don't. Hmm.

net/instaweb/rewriter/csp.cc
morlovich: Hmm, thought of it some more, and actually integration with redirects is quite nightmarish.
Basically the problem case is: different pages can have different CSPs, and our caching would be completely unaware of it.

net/instaweb/rewriter/csp_test.cc
morlovich: Adding some.... Very good point.

net/instaweb/rewriter/public/csp.h
```diff
@@ -42,29 +49,41 @@ class CspSourceExpression {
...
+    }
 
-    GoogleString scheme_part;  // doesn't include :
-    GoogleString host_part;
+    GoogleString scheme_part;  // doesn't include :, lowercased.
+    GoogleString host_part;    // lowercased.
     GoogleString port_part;
-    GoogleString path_part;
+    std::vector<GoogleString> path_part;  // normalized, separated by /
+    bool path_exact_match;
```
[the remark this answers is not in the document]
morlovich: Rewrote the comments here to give a hopefully better overview. Not a fan of lowercased_ names due to length.

net/instaweb/rewriter/csp.cc
oschaaf: If we could store a list of all the `Location:` headers we followed along with the final response, maybe we could check CSP at output time?

morlovich:
On Wed, Mar 15, 2017 at 10:29 AM, Otto van der Schaaf <
notifications@github.com> wrote:

> *@oschaaf* commented on this pull request.
> ------------------------------
>
> In net/instaweb/rewriter/csp.cc
> <https://github.com/pagespeed/mod_pagespeed/pull/1512#discussion_r106178918>
> :
>
> > +
> +  if (expr_port.empty()) {
> +    if (!HasDefaultPortForScheme(url)) {
> +      return false;
> +    }
> +  } else {
> +    // TODO(morlovich): Check whether the :80/:443 case is about effective
> +    // or explicit port.
> +    if (expr_port != "*"
> +        && expr_port != IntegerToString(url.EffectiveIntPort())
> +        && !(expr_port == "80" && url.EffectiveIntPort() == 443)) {
> +      return false;
> +    }
> +  }
> +
> +  if (!expr_path.empty()) {  // this would also be skipped for redirects
>
> If we could store a list of all the Location: headers we followed along
> with the final response, maybe we could check CSP at output time?
>
Yeah, that sounds workable. Would have to store them in metadata cache and
propagating that across the layers would be annoying and fiddly, but it
could work correctly.

morlovich:
Huibao: Could you take a look at this (and its successors); Josh seems to be way too busy.
CL roadmap:
https://github.com/pagespeed/mod_pagespeed/wiki/Design-Doc:-Brainstorming-PageSpeed-Optimization-Products-and-Content-Security-Policy#implementation-status

jmarantz (approved):

PR #1540, opened 2017-05-26
Consolidate external CSS, fix unauth. inlining test

oschaaf:
Move a css file to modpagespeed.com to avoid external changes to its response from breaking our tests

### DIFF
--- a/install/mod_pagespeed_test/unauthorized/inline_css.html
+++ b/install/mod_pagespeed_test/unauthorized/inline_css.html
@@ -1,6 +1,7 @@
 <html>
   <head>
     <title>inline_css example</title>
+    <!-- modpagespeed.com is not authorized in the tests -->
     <link rel="stylesheet" href="http://modpagespeed.com/testfiles/google-cse-default.css">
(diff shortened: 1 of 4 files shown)


install/mod_pagespeed_test/unauthorized/inline_css.html
```diff
@@ -1,7 +1,7 @@
 <html>
   <head>
     <title>inline_css example</title>
-    <link rel="stylesheet" href="http://cse.google.com/cse/style/look/default.css">
+    <link rel="stylesheet" href="http://modpagespeed.com/testfiles/google-cse-default.css">
```
jmarantz: I think this might defeat the purpose of this test, which is about the behavior of the system with resources on unauthorized domains.  But I guess it's OK because this isn't in an example that would be run with the main page on modpagespeed.com; it would be run from whatever instance of mod_pagespeed was being tested.  Maybe worth a comment though.

oschaaf: yes, my idea was that `modpagespeed.com` is just as unauthorized in the tests as `cse.google.com`. I'll add a comment

jmarantz (approved):

PR #1554, opened 2017-06-06
Debug builds may abort() when revalidating expired output resources

oschaaf:
As a workaround, remove the LOG(DFATAL) and replace it with a TODO.

https://github.com/pagespeed/mod_pagespeed/issues/1553

### DIFF
--- a/pagespeed/automatic/proxy_interface_test.cc
+++ b/pagespeed/automatic/proxy_interface_test.cc
@@ -386,23 +386,29 @@ TEST_F(ProxyInterfaceTest, TestNoDebugAbortAfterMoreThenOneYear) {
   GoogleString text;
   RequestHeaders request_headers;
   ResponseHeaders response_headers;
+  RewriteOptions* options = server_context()->global_options();
+  static const char* kUrl = "http://test.com/flush_subresources.html";
+
+  options->ClearSignatureForTesting();
+  options->EnableFilter(RewriteOptions::kExtendCacheCss);
+  options->DisableFilter(RewriteOptions::kInlineCss);
+  server_context()->ComputeSignature(options);
+
   response_headers.Add(HttpAttributes::kContentType,
-		       kContentTypeHtml.mime_type());
+                       kContentTypeHtml.mime_type());
   response_headers.SetStatusAndReason(HttpStatus::kOK);
 
-  GoogleString url = "http://test.com/flush_subresources.html"
-    "?PageSpeedFilters=+extend_cache_css,-inline_css";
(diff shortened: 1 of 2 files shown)


pagespeed/automatic/proxy_interface_test.cc
```diff
@@ -381,6 +381,46 @@ TEST_F(ProxyInterfaceTest, LoggingInfo) {
...
+// Regression test for https://github.com/pagespeed/mod_pagespeed/issues/1553
+TEST_F(ProxyInterfaceTest, TestNoDebugAbortAfterMoreThenOneYear) {
+  GoogleString text;
+  RequestHeaders request_headers;
+  ResponseHeaders response_headers;
+  response_headers.Add(HttpAttributes::kContentType,
+		       kContentTypeHtml.mime_type());
+  response_headers.SetStatusAndReason(HttpStatus::kOK);
+
+  GoogleString url = "http://test.com/flush_subresources.html"
```
jmarantz: static const char kUrl[] =

pagespeed/automatic/proxy_interface_test.cc
```diff
@@ -381,6 +381,46 @@ TEST_F(ProxyInterfaceTest, LoggingInfo) {
...
+    "?PageSpeedFilters=+extend_cache_css,-inline_css";
```
jmarantz: would prefer setting this in options, e.g. with this pattern elsewhere in the file:

  RewriteOptions* options = server_context()->global_options();
  options->ClearSignatureForTesting();
  options->EnableFilter(RewriteOptions::kExtendCacheCss);
  options->DisableFilter(RewriteOptions::kInlineCss);
  server_context()->ComputeSignature(options);
  
Then you can omit the query-params from the URL, and this will behave more like normal usage.  Admittedly this takes more lines, but I think it's more important to mimic live usage.

pagespeed/automatic/proxy_interface_test.cc
```diff
@@ -381,6 +381,46 @@ TEST_F(ProxyInterfaceTest, LoggingInfo) {
...
+  FetchFromProxy(url,
+                 request_headers,
+                 true,  /* expect_success */
+                 &text,
+                 &response_headers,
+                 false  /* proxy_fetch_property_callback_collector_created */);
+
+  SetTimeMs(MockTimer::kApr_5_2010_ms + 2 * Timer::kYearMs);
+
+  // This second fetch would end up with VLOG(DFATAL)
```
jmarantz: ....prior to changing that to a TODO :)

jmarantz (approved):

jmarantz (approved):

PR #1558, opened 2017-06-09
load test tweaks

oschaaf:
A couple of things I needed to tweak when trying to run the load tests

### DIFF
--- a/devel/mps_load_test.sh
+++ b/devel/mps_load_test.sh
@@ -215,11 +215,11 @@ make -j8 CONF=$compile_mode apache_trace_stress_test_server \
 
 # If a custom .so got specified, install it.
 if [[ -n "$custom_so" ]]; then
-  install -c $custom_so $HOME/apache2/modules/mod_pagespeed.so
+  install -c $custom_so $APACHE_DEBUG_ROOT/modules/mod_pagespeed.so
 fi
 
 if [[ -n "$custom_so24" ]]; then
-  install -c $custom_so24 $HOME/apache2/modules/mod_pagespeed_ap24.so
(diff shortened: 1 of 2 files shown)


devel/mps_load_test.sh
```diff
@@ -215,11 +215,11 @@ make -j8 CONF=$compile_mode apache_trace_stress_test_server \
 
 # If a custom .so got specified, install it.
 if [[ -n "$custom_so" ]]; then
-  install -c $custom_so /usr/local/apache2/modules/mod_pagespeed.so
+  install -c $custom_so $HOME/apache2/modules/mod_pagespeed.so
```
jmarantz: Is $HOME always right?  For that to go /home/jmarantz/apache2/...

maybe better to have APACHE_DIR which defaults to /usr/local/apache2 if not already defined as an env var.

oschaaf: Well, the release process doc goes:

```
Please load-test using multiple Apache version & MPMs 
..... 
These can all be built using install/build_development_apache.sh, which builds a new ~/apache2
```

So it surprised me that `devel/mps_load_test.sh` pointed to a location that didn't exist on my machine. If we only load test in the release flow, this change works..

jmarantz: Sorry -- we already have the variable I was talking about.  Use $APACHE_DEBUG_ROOT here, defined above as I described.

oschaaf: Nice, done in a971e89cc4c375ab7b00b2be16684af65d2cc941

jmarantz (approved):

jmarantz (approved):
You could while you are in here also quote the other usage of APACHE_DEBUG_ROOT but it's up to you.

jmarantz (approved):

PR #1559, opened 2017-06-09
mps-load-test: Specify BUILDTYPE=Debug

oschaaf:
On a clean checkout I noticed the tests failed with this on the terminal:

```
Built mod_pagespeed successfully:
ls -l /mps/out/Debug/libmod_pagespeed.so
ls: cannot access /mps/out/Debug/libmod_pagespeed.so: No such file or directory
```

- Specify BUILDTYPE=Debug, to build libmod_pagespeed.so at the
expected location
- Remove obsolete comment/TODO
- Add TODO

### DIFF
--- a/devel/Makefile
+++ b/devel/Makefile
@@ -125,8 +125,6 @@ DEVEL_DIR = $(GIT_SRC)/devel
 INSTALL_DATA_DIR = $(GIT_SRC)/install
 
 # Where binaries go.
-GIT_RELEASE_BIN = $(GIT_SRC)/out/Release
-GIT_DEBUG_BIN = $(GIT_SRC)/out/Debug
 GIT_BIN = $(GIT_SRC)/out/$(BUILDTYPE)
 
 # GYP_MANIFEST holds a cached copy of the gyp files in this client.  When we
@@ -296,19 +294,19 @@ apache_debug : $(HOOKS_STAMP)
 	@echo "building Apache module $@ ..."
 	cd $(GIT_SRC) && $(MAKE)
 	@echo Built mod_pagespeed successfully:
-	ls -l $(GIT_DEBUG_BIN)/libmod_pagespeed.so
+	ls -l $(GIT_BIN)/libmod_pagespeed.so
 	@echo To install, type
-	@echo "   " cp $(GIT_DEBUG_BIN)/libmod_pagespeed.so \
+	@echo "   " cp $(GIT_BIN)/libmod_pagespeed.so \
 		$(APACHE_DEBUG_ROOT)/modules/mod_pagespeed.so
 
 apache_debug_psol : apache_debug
 	cd $(GIT_SRC)/pagespeed/automatic && $(MAKE) \
 		MOD_PAGESPEED_ROOT=$(GIT_SRC) \
-		OUTPUT_DIR=$(GIT_DEBUG_BIN) \
+		OUTPUT_DIR=$(GIT_BIN) \
 		BUILDTYPE=Debug \
 		CXXFLAGS=$(CXXFLAGS) \
 		all
-	@echo Built PSOL successfully under in $(GIT_DEBUG_BIN)
+	@echo Built PSOL successfully under in $(GIT_BIN)
 
 .PHONY : submodule_update
 submodule_update:
@@ -561,9 +559,9 @@ TMP_PREFIX = /tmp/mod_pagespeed.$(USER).install
 internal_release_test :
 	cd $(GIT_SRC) && \
 	  BUILDTYPE=Release install/run_program_with_ext_caches.sh \
-	    $(GIT_RELEASE_BIN)/mod_pagespeed_test  $(TEST_ARG) && \
+	    $(GIT_BIN)/mod_pagespeed_test  $(TEST_ARG) && \
 	  BUILDTYPE=Release install/run_program_with_ext_caches.sh \
-	    $(GIT_RELEASE_BIN)/pagespeed_automatic_test $(TEST_ARG)
+	    $(GIT_BIN)/pagespeed_automatic_test $(TEST_ARG)
 
 # Configuration root for Apache file-system and cache directories, to
 # be written into config file.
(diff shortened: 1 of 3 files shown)


devel/mps_load_test.sh
```diff
@@ -208,7 +202,8 @@ cd "$src/devel"
 
 # Build a version of mod_pagespeed with all optimizations enabled, but with
 # a build that includes DCHECKs.
-make -j8 CONF=$compile_mode apache_trace_stress_test_server \
```
morlovich: IIRC CONF= was supposed to set BUILDTYPE? (see comments on top of devel/Makefile)
In particular we want to loadtest OptDebug, not pure debug.

oschaaf: Ah, CONF=OptDebug overrides BUILDTYPE to equal Release in devel/Makefile. Let me debug this some more.

oschaaf: I'm a bit confused. There's this in `devel/Makefile`:

```
apache_trace_stress_test_server :
        $(MAKE) apache_trace_stress_test_prepare
        $(MAKE) apache_debug_install
        $(MAKE) apache_debug_start
``` 

That is always going to expect libmod_pagespeed.so in `out/Debug/`, right? But the `libmod_pagespeed.so` we build will be ending up in `out/Release/`. So if the script succeeds, we're probably testing the wrong lib, but on a fresh checkout/build the script will fail.

morlovich: Hmm, I think there might be inconsistency between apache_debug_install's current and historical meanings that only half-ways works.  Roughly it used to be the apache_debug_ just meant the apache in APACHE_DEBUG_ROOT, with whatever BUILDTYPE= you would pass, and IIRC the apache_release_ targets touched the system apache (probably, since I never used the apache_release_... ones). I am pretty sure the apache_trace_stress_test_server target was written with this "historical" interpretation.

In fact the actual install portion of apache_debug_install works just fine:

apache_debug_install : apache_debug
	install -c $(GIT_BIN)/libmod_pagespeed.so \
	           $(APACHE_DEBUG_MODULES)/mod_pagespeed.so

GIT_BIN = $(GIT_SRC)/out/$(BUILDTYPE)
(Notice also:
 APACHE_DEBUG_MODULES = $(APACHE_DEBUG_ROOT)/modules
 ... which uses the 'old' meaning of debug)

OTOH:
apache_debug : $(HOOKS_STAMP)
	@echo "building Apache module $@ ..."
	cd $(GIT_SRC) && $(MAKE)
	@echo Built mod_pagespeed successfully:
	ls -l $(GIT_DEBUG_BIN)/libmod_pagespeed.so
	@echo To install, type
	@echo "   " cp $(GIT_DEBUG_BIN)/libmod_pagespeed.so \
		$(APACHE_DEBUG_ROOT)/modules/mod_pagespeed.so

... where GIT_DEBUG_BIN is build-type specific:
GIT_DEBUG_BIN = $(GIT_SRC)/out/Debug

One option might be to basically kill every distinction that uses GIT_DEBUG_BIN vs. GIT_RELEASE_BIN, but fixing all the dependencies in scripts/Makefiles/docs might be unreasonably messy.

oschaaf: Thanks, that helps a lot with understanding. I'm having a go at killing `GIT_DEBUG_BIN` and `GIT_RELEASE_BIN`... 

One more observation: If you run `loadtest_collect_corpus.sh` on a clean checkout,  apache responds with every request with the stock hello world response (it works!).
If I run `apache_debug_smoke_test` and proceed to  run `loadtest_collect_corpus.sh` I get a real corpus.

oschaaf: @morlovich done - see e5b14ad6b4efbd8f668692cc8d5eeb24fa0741c9.
So we only have `GIT_BIN` now.
This seems to work fine for me, do you foresee any issues with it?

oschaaf:
one more note on `mps_load_test.sh`:
At some point the script goes ` # Print some interesting statistics from the server`. AFAICT, the stats it greps are never shown on the output (the output from trace_stress_test.sh does show up).

oschaaf:
whoops, I was wrong, the stats *do* show up, sorry!

devel/loadtest_collect/loadtest_collect_corpus.sh
```diff
@@ -55,6 +55,7 @@ make apache_debug_stop
 sed -e "s^#HOME^$HOME^" -e "s^#SLURP_DIR^$SLURP_DIR^" \
   -e "s^#LOG_PATH^$LOG_PATH^" \
   < loadtest_collect/loadtest_collect.conf > ~/apache2/conf/pagespeed.conf
+cat ~/apache2/conf/pagespeed.conf
```
morlovich, line 58: Did you mean to leave this in?

devel/mps_load_test.sh
```diff
@@ -208,6 +202,7 @@ cd "$src/devel"
 
 # Build a version of mod_pagespeed with all optimizations enabled, but with
 # a build that includes DCHECKs.
+# TODO(oschaaf): can we skip this step if we have a custom .so?
```
morlovich, line 205: It sets up the config as well, so you would want to at least use apache_trace_stress_test_prepare

devel/Makefile
```diff
@@ -561,9 +559,9 @@ TMP_PREFIX = /tmp/mod_pagespeed.$(USER).install
 internal_release_test :
 	cd $(GIT_SRC) && \
 	  BUILDTYPE=Release install/run_program_with_ext_caches.sh \
-	    $(GIT_RELEASE_BIN)/mod_pagespeed_test  $(TEST_ARG) && \
+	    $(GIT_BIN)/mod_pagespeed_test  $(TEST_ARG) && \
```
morlovich, line 562: hmm, does this actually work? I think nothing overrides BUILDTYPE within Make for this target (unlike in internal_release_build where it passes it to MAKE)

oschaaf:
Closing this for now, the siege tests adressed their part, I'll get back to this later

PR #1569, opened 2017-06-20
Doc: remove "beta" after verifying build from source docs

oschaaf:


morlovich:
LGTM, given
NPS_VERSION=1.12.34.2-stable
in the doc above, and that matching the dir in the tarball

oschaaf:
thanks, deployed!

PR #1571, opened 2017-06-27
Not adding type attribute in script tag, adding attribute when pedantic filter is enabled : Bug fix #1207

### LINKED ISSUE #1207: Don't emit type=text/javascript on script tags
Hi,
I noticed mod_pagespeed includes a new "script" tag inside "head". Which is normal, but in this case the head already had a "script" tag:

```
<!DOCTYPE html><html xmlns="http://www.w3.org/1999/xhtml" xmlns:g="http://base.google.com/ns/1.0" xmlns:og="http://ogp.me/ns#" xmlns:fb="http://ogp.me/ns/fb#" xml:lang=en lang=en>
<head>
<script type='text/javascript'>window.mod_pagespeed_start = Number(new Date());</script><script>var FileAPI={
```

Maybe there could be a way to merge these tags? Although their declaration is not strictly identical, I have heard that for certain doctypes, the "type" attribute of the "script" tag is unnecessary. Maybe stripping this attribute could be another "less bytes" optimization for pagespeed, and in this case merging two tags as well?
Thanks,

ashishk-1:
- In add_instrumentation_filter.cc, I have added pedantic filter check when adding type attribute to script tag.

- In common_filter.cc file, function AddJsToElement, adds type attribute based on HTML5 version check. I have added 'OR' condition of pedantic filter enabled check. So if it **not HTML5 OR pedantic filter enabled**, it will **add type attribute**.
Is this check appropriate ? Or only pedantic filter check is sufficient?
Let me know your thoughts on this.

### DIFF
--- a/net/instaweb/rewriter/add_instrumentation_filter.cc
+++ b/net/instaweb/rewriter/add_instrumentation_filter.cc
@@ -101,11 +101,11 @@ void AddInstrumentationFilter::AddHeadScript(HtmlElement* element) {
     added_head_script_ = true;
     // TODO(abliss): add an actual element instead, so other filters can
     // rewrite this JS
-    HtmlCharactersNode* script = NULL;
-    if(driver()->options()->Enabled(RewriteOptions::kPedantic)){
-        script = driver()->NewCharactersNode(NULL, kHeadScriptPedantic);
+    HtmlCharactersNode* script = nullptr;
+    if (driver()->options()->Enabled(RewriteOptions::kPedantic)) {
+        script = driver()->NewCharactersNode(nullptr, kHeadScriptPedantic);
     } else {
-        script = driver()->NewCharactersNode(NULL, kHeadScriptNonPedantic);
+        script = driver()->NewCharactersNode(nullptr, kHeadScriptNonPedantic);
     }
     driver()->InsertNodeBeforeCurrent(script);
     instrumentation_script_added_count_->Add(1);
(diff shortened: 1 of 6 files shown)


net/instaweb/rewriter/add_instrumentation_filter.cc
```diff
@@ -95,7 +101,12 @@ void AddInstrumentationFilter::AddHeadScript(HtmlElement* element) {
     added_head_script_ = true;
     // TODO(abliss): add an actual element instead, so other filters can
     // rewrite this JS
-    HtmlCharactersNode* script = driver()->NewCharactersNode(NULL, kHeadScript);
+    HtmlCharactersNode* script = NULL;
+    if(driver()->options()->Enabled(RewriteOptions::kPedantic)){
```
jmarantz: We use a space between "if" and "(driver)", and before the "{"

https://google.github.io/styleguide/cppguide.html#Conditionals

net/instaweb/rewriter/add_instrumentation_filter.cc
```diff
@@ -95,7 +101,12 @@ void AddInstrumentationFilter::AddHeadScript(HtmlElement* element) {
...
+        script = driver()->NewCharactersNode(NULL, kHeadScriptPedantic);
```
jmarantz: s/NULL/nullptr/ in new code.  The codebase was written pre-cxx11 so there are some NULLs in this file.  It's up to you if you want to fix those while you are in here.

jmarantz (approved):
nice!  just some trivial formatting comments.

Also I think some of your lines may be >80 chars.

jmarantz (approved):

PR #1584, opened 2017-07-05
Update deps: apr, apr-util, and serf

oschaaf:
* Update deps: apr, aprutil, serf

apr:  1.5.1 -> 1.6.2
aprutil:  1.5.4 -> 1.6.0
serf: 1.3.8 -> 1.3.9

* Add libexpat1-dev (apr 1.6.0 unbundled it)

### DIFF
--- a/install/ubuntu/install_required_packages.sh
+++ b/install/ubuntu/install_required_packages.sh
@@ -36,7 +36,7 @@ if [ $# -ne 0 ]; then
 fi
 
 binary_packages=(subversion apache2 g++ gperf devscripts fakeroot git-core
-  zlib1g-dev wget curl net-tools rsync ssl-cert psmisc)
+  zlib1g-dev wget curl net-tools rsync ssl-cert psmisc libexpat1-dev)
 src_packages=()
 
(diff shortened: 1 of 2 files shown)


morlovich:
Could you please include the version info in the top-level description (like on your commits does)

So what makes this non-trivial is that we compile against those headers, but --- in Apache only, not even mod_pagespeed unit tests --- link against the version in the httpd binary. I think they've been keeping things ABI stable, thankfully, so it's not a libpng situation and as long as we don't use the new stuff we're fine.

oschaaf:
re: "Could you please include the version info in the top-level description (like on your commits does)"
Upon merge, this would get squashed into a single commit with a description like:

```
* Update deps: apr, aprutil, serf

apr:  1.5.1 -> 1.6.2
aprutil:  1.5.4 -> 1.6.0
serf: 1.3.8 -> 1.3.9

* Add libexpat1-dev (apr 1.6.0 unbundled it)
```

Does that look good to you?

morlovich:
LGTM.

PR #1588, opened 2017-07-06
Deflake: attempt to deflake ipro_noop

oschaaf:


oschaaf:
Some context here, I observed a flake in this test, and I can only attribute that to fetches originating from `fetch_until` lagging. My thought is that waiting for serf_fetch_active_count to drain should stabilize the last check:
```
check [ $NUM_FETCHES_URL2 = $NUM_FETCHES_URL1 ]
```

morlovich:
I am confused, isn't that commented out? Or was it commented out for lack of better fix? In the latter case the change makes sense, but perhaps a function would be appropriate...

oschaaf:
@morlovich I was confused too, but: I think you may have to expand the diff downwards: it doesn't initially show the line I mentioned in my second comment, though the one the initial diff view ends with looks pretty similar at a glance.. Including one more line:

```
  #check [ $NUM_REWRITES_URL2 = $NUM_REWRITES_URL1 ]
 check [ $NUM_FETCHES_URL2 = $NUM_FETCHES_URL1 ]
```
Does that clarify the change?

I'll add the function to deduplicate the waiting for quiescence code.

pagespeed/system/system_tests/ipro_noop.sh
```diff
@@ -48,9 +48,20 @@ URL1="${URL}&PageSpeedNoop=$RANDOM1"
 URL2="${URL}&PageSpeedNoop=$RANDOM2"
 fetch_until "$URL1" "grep -c image/webp" 1 --save-headers
 #NUM_REWRITES_URL1=$(scrape_stat image_rewrites)
+echo -n Waiting for quiescence by checking serf_fetch_active_count ...
+# The last check in this test was observed to flake.Let's see if waiting
```
jmarantz: Add space after "flake."

pagespeed/system/system_tests/ipro_noop.sh
```diff
@@ -48,9 +48,20 @@ URL1="${URL}&PageSpeedNoop=$RANDOM1"
...
+# for quiescence here stabilizes it.
+while [ $(scrape_stat serf_fetch_active_count) -gt 0 ]; do
```
jmarantz: It would be good to keep a count of active rewrite drivers as well for this purpose.  Serf is only one of the things that can finish asynchronously; slow rewrites can as well.

This can be a TODO as this will probably help and I don't want to block you.  And keeping an active rewrite_count is not necessarily trivial.

jmarantz (approved):

PR #1600, opened 2017-07-27
Dont inline css from style link element in body when pedantic filter enabled. This is to maintain w3c validation of page. Bug fix #1153

### LINKED ISSUE #1153: pedantic filter should prevent inlining CSS into the body
Automatic CSS inlining leaves `<style>` tag inside `<body>`, making the pages invalid.

ashishk-1:
When pedantic filter is enabled, style link element in body is not inlined. When move_css_to_head filter is enabled, inlining is allowed because css is moved to head which is valid as per w3c standard.

### DIFF
--- a/net/instaweb/rewriter/css_inline_filter.cc
+++ b/net/instaweb/rewriter/css_inline_filter.cc
@@ -98,7 +98,7 @@ CssInlineFilter::CssInlineFilter(RewriteDriver* driver)
     : CommonFilter(driver),
       id_(RewriteOptions::kCssInlineId),
       size_threshold_bytes_(driver->options()->css_inline_max_bytes()),
-      in_body_(false) {
+      inlining_not_pedantically_spec_(false) {
   Statistics* stats = server_context()->statistics();
   num_css_inlined_ = stats->GetVariable(kNumCssInlined);
 }
@@ -113,8 +113,11 @@ void CssInlineFilter::StartDocumentImpl() {
 CssInlineFilter::~CssInlineFilter() {}
 
 void CssInlineFilter::StartElementImpl(HtmlElement* element) {
-  if (element->keyword() == HtmlName::kBody) {
-    in_body_ = true;
+
+  // w3c incompatible when css inlining in body with pedantic enabled
+  if (element->keyword() == HtmlName::kBody &&
+      driver()->options()->Enabled(RewriteOptions::kPedantic)) {
+    inlining_not_pedantically_spec_ = true;
   }
 }
 
@@ -122,10 +125,6 @@ void CssInlineFilter::EndElementImpl(HtmlElement* element) {
   // Don't inline if the CSS element is under <noscript>.
   if (noscript_element() != NULL) {
     return;
-  }
-
-  if (element->keyword() == HtmlName::kBody) {
-    in_body_ = false;
   }
 
   HtmlElement::Attribute* href = NULL;
@@ -152,11 +151,10 @@ void CssInlineFilter::EndElementImpl(HtmlElement* element) {
     }
 
     // Dont inline if style <link> element is in html body AND
-    // move_css_to_head is not enabled. Pedantic used for html compatibility.
+    // move_css_to_head is not enabled.
     // This is to maintain w3c validation since style element is
     // not recommended in html body. Issue fix #1153.
-    if (in_body_ &&
-        driver()->options()->Enabled(RewriteOptions::kPedantic) &&
+    if (inlining_not_pedantically_spec_ &&
         !driver()->options()->Enabled(RewriteOptions::kMoveCssToHead)) {
       driver()->InsertDebugComment(
           "CSS not inlined because style link element in html body", element);
--- a/net/instaweb/rewriter/css_inline_filter.cc
+++ b/net/instaweb/rewriter/css_inline_filter.cc
@@ -97,8 +97,7 @@ class CssInlineFilter::Context : public InlineRewriteContext {
 CssInlineFilter::CssInlineFilter(RewriteDriver* driver)
     : CommonFilter(driver),
       id_(RewriteOptions::kCssInlineId),
-      size_threshold_bytes_(driver->options()->css_inline_max_bytes()),
-      inlining_not_pedantically_spec_(false) {
+      size_threshold_bytes_(driver->options()->css_inline_max_bytes()) {
   Statistics* stats = server_context()->statistics();
   num_css_inlined_ = stats->GetVariable(kNumCssInlined);
 }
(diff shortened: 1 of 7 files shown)


net/instaweb/rewriter/css_inline_filter.cc
```diff
@@ -111,11 +112,22 @@ void CssInlineFilter::StartDocumentImpl() {
...
+}
+
 void CssInlineFilter::EndElementImpl(HtmlElement* element) {
   // Don't inline if the CSS element is under <noscript>.
   if (noscript_element() != NULL) {
     return;
   }
+
+  if (element->keyword() == HtmlName::kBody) {
+    in_body_ = false;
```
oschaaf: Do you see any direct benefit in setting `in_body_` to `false` again here?
I can see that doing this ensures that the value keeps reflecting the name (`in_body_`), but if that is all, maybe we should rename `in_body_` to `inlining_not_pedantically_spec_` and remove this code?

oschaaf (commented):
Thanks! Mostly looks good to me, one question.

net/instaweb/rewriter/css_inline_filter.cc
ashishk-1: In case <body> element comes before <head>, so thought of making in_body_ to false in EndElementImpl(). But actually for w3c compatibility, this original structure itself will not be compatible. So we can remove this.  
I can also rename this flag as you mentioned.

ashishk-1: In above post, I meant if body element comes before head element.

(formatting of angulat brackets removed element name)

pagespeed/system/system_tests/inline_css_link_element_in_body.sh
```diff
@@ -0,0 +1,32 @@
...
+# AND pedantic filter is enabled
+$WGET -O $WGET_OUTPUT $TEST_ROOT/inline_style_link_in_body/inlining_style_link_body.html\
+?PageSpeedFilters=pedantic,inline_css
+# checking for no inlining of css
+check [ $(grep -c "href=\"style.css\"" $WGET_OUTPUT) = 1 ]
+
+# css inlined and moved to head
+URL="http://localhost:8080/mod_pagespeed_test/inline_style_link_in_body/inlining_style_link_body.html?PageSpeedFilters=pedantic,inline_css,move_css_to_head"
+# checking contents of inlined css
+# check [ $(grep -c ".foo" $WGET_OUTPUT) = 1 ]
```
oschaaf: stray comment?

pagespeed/system/system_tests/inline_css_link_element_in_body.sh
```diff
@@ -0,0 +1,32 @@
...
+
+# css not inlined because style link element in body
+# AND pedantic filter is enabled
+$WGET -O $WGET_OUTPUT $TEST_ROOT/inline_style_link_in_body/inlining_style_link_body.html\
+?PageSpeedFilters=pedantic,inline_css
+# checking for no inlining of css
+check [ $(grep -c "href=\"style.css\"" $WGET_OUTPUT) = 1 ]
+
+# css inlined and moved to head
+URL="http://localhost:8080/mod_pagespeed_test/inline_style_link_in_body/inlining_style_link_body.html?PageSpeedFilters=pedantic,inline_css,move_css_to_head"
```
oschaaf: Use `$TEST_ROOT` instead of  `localhost:8080`?

oschaaf (changes requested):
Two small requests for the system test, but other then that this looks good to me.

oschaaf (approved):

oschaaf:
@jmarantz are you ok with this being merged?

net/instaweb/rewriter/css_inline_filter.cc
```diff
@@ -97,7 +97,8 @@ class CssInlineFilter::Context : public InlineRewriteContext {
 CssInlineFilter::CssInlineFilter(RewriteDriver* driver)
     : CommonFilter(driver),
       id_(RewriteOptions::kCssInlineId),
-      size_threshold_bytes_(driver->options()->css_inline_max_bytes()) {
+      size_threshold_bytes_(driver->options()->css_inline_max_bytes()),
+      inlining_not_pedantically_spec_(false) {
```
jmarantz: This is actually a document-state bit, which I didn't realize until I saw where it was spec.  So you need to init it in StartDocumentImpl, as the same CssInlineFilter class instance can be re-used for multiple documents (we recycle drivers for compatible configurations).

Also I think I would call the bit has_pedantic_inline_violations_.

net/instaweb/rewriter/css_inline_filter.cc
```diff
@@ -111,11 +112,21 @@ void CssInlineFilter::StartDocumentImpl() {
 
 CssInlineFilter::~CssInlineFilter() {}
 
+void CssInlineFilter::StartElementImpl(HtmlElement* element) {
+
+  // w3c incompatible when css inlining in body with pedantic enabled
+  if (element->keyword() == HtmlName::kBody &&
+      driver()->options()->Enabled(RewriteOptions::kPedantic)) {
+    inlining_not_pedantically_spec_ = true;
```
jmarantz: I don't understand what's being tracked here.  This function gets called for every document element when inlining is on, and there will always be a 'body' element.  I think this is tracking whether we are inside the 'body' in which case it might be less confusing to call this variable in_body_ (contradicting my suggestion above).

net/instaweb/rewriter/css_inline_filter.cc
```diff
@@ -138,6 +149,18 @@ void CssInlineFilter::EndElementImpl(HtmlElement* element) {
           "CSS not inlined because media does not match screen", element);
       return;
     }
+
+    // Dont inline if style <link> element is in html body AND
+    // move_css_to_head is not enabled.
+    // This is to maintain w3c validation since style element is
+    // not recommended in html body. Issue fix #1153.
+    if (inlining_not_pedantically_spec_ &&
```
jmarantz: shouldn't this be ANDed with Enabled(RewriteOptions::kPedantic)?

jmarantz:
I left some comments.  I think this has some corner-cases that need to be
addressed:
  - behavior with 'pedantic' off
  - behavior when the same RewriteDriver object processes a second document

net/instaweb/rewriter/css_inline_filter.cc
ashishk-1: I am tracking whether we are in body with pedantic filter enabled. I can change this variable to in_body_ and will shift pedantic filter enable check to EndElementImpl() function, so that it will become less confusing. I will also init variable in StartDocumentImpl.

net/instaweb/rewriter/css_inline_filter.cc
ashishk-1: I ANDed pedantic filter enable check in StartElementImpl where I was checking whether we are in body. 
Now I am changing the variable to in_body_ and will also move pedantic filter enable check in endElementImpl().

ashishk-1:
>   - behavior with 'pedantic' off

when pedantic filter is off, inlining_not_pedantically_spec_(now changing to in_body_) will be false always and will inline the css even if style link element is there in html body. I will add one more unit and system test for this.

> - behavior when the same RewriteDriver object processes a second document

I will add the test of multiple page loading in single test.

oschaaf:
any objections against merging this?

net/instaweb/rewriter/css_inline_filter.cc
```diff
@@ -107,15 +107,24 @@ void CssInlineFilter::InitStats(Statistics* statistics) {
 }
 
 void CssInlineFilter::StartDocumentImpl() {
+    in_body_ = false;
```
jmarantz, line 111: also init this in the constructor (I know this is redundant but it just seems cleaner to make sure it doesn't ever have garbage in it).  I think in practice this won't matter.

jmarantz (approved):

PR #1605, opened 2017-07-28
Skeleton of CSP integration in mod_pagespeed.

morlovich:
Hooks up CSP with RewriteDriver and ScanFilter and an off switch. Makes one filter use it.

### DIFF
--- a/net/instaweb/rewriter/responsive_image_filter.cc
+++ b/net/instaweb/rewriter/responsive_image_filter.cc
@@ -434,7 +434,7 @@ void ResponsiveImageSecondFilter::EndDocument() {
   if (zoom_filter_enabled_ && srcsets_added_ && !driver()->is_amp_document()) {
     if (driver()->IsLoadPermittedByCsp(
           CspDirective::kScriptSrc, responsive_js_url_)) {
-      HtmlElement* script = driver()->NewElement(NULL, HtmlName::kScript);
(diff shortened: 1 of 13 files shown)


net/instaweb/rewriter/responsive_image_filter.cc
```diff
@@ -432,9 +432,17 @@ void ResponsiveImageSecondFilter::Cleanup(
 
 void ResponsiveImageSecondFilter::EndDocument() {
   if (zoom_filter_enabled_ && srcsets_added_ && !driver()->is_amp_document()) {
-    HtmlElement* script = driver()->NewElement(NULL, HtmlName::kScript);
-    driver()->AddAttribute(script, HtmlName::kSrc, responsive_js_url_);
-    InsertNodeAtBodyEnd(script);
+    if (driver()->IsLoadPermittedByCsp(
+          CspDirective::kScriptSrc, responsive_js_url_)) {
+      HtmlElement* script = driver()->NewElement(NULL, HtmlName::kScript);
```
jmarantz: nullptr

oschaaf:
great, lgtm!

jmarantz (approved):
at some point will there be a system test?

net/instaweb/rewriter/responsive_image_filter.cc
morlovich: Done; doesn't seem like I have an integration test in any of the outstanding CLs, though.

PR #1612, opened 2017-08-08
Don't use signed here, hash computation can overflow.

morlovich:
(And signed overflow is undefined behavior)


jmarantz (approved):

morlovich:
The failure seems like trouble talking to git.apache.org, so I am going ahead.

PR #1613, opened 2017-08-08
Giflib: use updated submodule (4.1.6 -> 5.1.4)

oschaaf:


morlovich:
LGTM --- given that 5.x has thread safety fixes, and we already support it, seems like a good idea to me.

Issue #1572, opened 2017-06-27
CSS minifier breaks unicode-range values


peterbleickardt:
What steps will reproduce the problem?
1. Activate CSS minification in mod_pagespeed
2. Include a `unicode-range` CSS declaration in a stylesheet: `unicode-range: U+0400-045F, U+0490-0491, U+04B0-04B1, U+2116`

What is the expected output?
`unicode-range:U+0400-045F, U+0490-0491, U+04B0-04B1, U+2116`

What do you see instead?
The CSS minification output breaks the value, placing a space between the 'U' and '+':

`unicode-range:U +0400-045F , U +0490-0491 , U +04B0-04B1 , U +2116`

This prevents the browser from seeing the proper value, therefore always downloading the font even if the intended range of characters is not used.

What version of the product are you using?
X-Mod-Pagespeed: 1.11.33.4-0

Which OS? Which version of Apache? Which MPM?
CentOS, Apache/2.2.27 (prefork)

jmarantz:
Thanks for the report.   For quick workaround, I suggest using ModPagespeedDisallow to avoid processing the CSS file with the unicode characters.  Hopefully this (a) doesn't affect all of them and (b) maybe you could split it just the rules that have this.

I'll check into what the status of this is.  A few things are fixed in 1.12 (that's now stable) however this one doesn't ring a bell.

Issue #1538, opened 2017-05-17
CSS Minification - 0 followed by a unit is rewritten to 0 without a unit.


dhaupin:
I ran into this today with mod_pagespeed 1.11.33.4-0 that is packaged in cPanel experimental repo.

When using CSS minifier, 0 followed by a unit, such as `0px`, is re-written into `0` without a unit. This breaks some CSS directives that always require a unit. Example is a dynamic construction of CSS that drops a `0px` into a `calc()` when an offset is not required.

Example:
`left: calc(50% - $offset);`

Since its not always used, `$offset` is set to `0px` by default, which renders as:
`left: calc(50% - 0px);`

Upon minification, Pagespeed truncates the `px` resulting in:
`left: calc(50% - 0);`

This is invalid and causes errors. Besides Pagespeed, this issue seems to be found in other minification applications as well. Is there any possibility to always include the unit of measurement? The spec claims unit is optional, so perhaps there should be a config for it rather than assuming "no unit" on rewrite: https://www.w3.org/TR/css3-values/#lengths

Thanks.

jmarantz:
I think this is fixed in mod_pagespeed 1.12.  The source-code was fixed here:

https://github.com/pagespeed/mod_pagespeed/commit/7c30b7ec44f7534f984394f710c53c6ea0e8bff8

Can you switch to the beta channel?  We should be promoting this to stable soon.

dhaupin:
Nice, that is a similar issue. Don't have much time this eve to check further, but that looks like that commit is dealing with percents and seconds. Its def the same kinda hiccup, however this is every unit that CSS requires specifically in areas like `calc()`. Like transition, its invalid to use a non-unit 0 and invalids the style rule.

As far as the beta, I personally would be down, but in this case, we are relying on cPanel experimental branch for "EA4", which pulls down a stable release they have gone through. Wish i had time to go through and try to figure out a pull, but its super busy lately. Maybe i can do a vanilla beta on a new instance without cPanel.

Anyways, thanks for your time looking into this. Appreciate it.

Suranex:
We use mod_pagespeed version 1.12.34.2-0 on Debian and have still this issue. calc(100% - 0px); is optimized to calc(100% - 0).

PR #1616, opened 2017-08-18
CSS minification issues fix. Issue #1538 and #1572 fix.

ashishk-1:
- UnitsRequiredForValueZero() function from css_minify.cc was in the anonymous namespace.I have moved it to default(net_instaweb) namespace.The reason being I need to use member variable(in_css_calc_function_) in this function to check whether unit is required or not.
I could have checked this variable by ORing here https://github.com/pagespeed/mod_pagespeed/blob/master/net/instaweb/rewriter/css_minify.cc#L465
But I find it more appropriate inside UnitsRequiredForValueZero() function.

- In CssMinify::Minify(const Css::Declaration& declaration) function, there is switch case where css property element gets resolved. I could not find appropriate enum for unicode-range property in that (third party css parser implementation). Else the implementation would have been much cleaner.
There is TODO for implementing parser for unicode-range descriptor.
https://github.com/pagespeed/mod_pagespeed/blob/master/third_party/css_parser/src/webutil/css/parser_unittest.cc#L2424

### DIFF
--- a/install/mod_pagespeed_test/css_minify_calc_function_value_zero.html
+++ b/install/mod_pagespeed_test/css_minify_calc_function_value_zero.html
@@ -1,0 +1,5 @@
+<html>
+  <head>
+     <style property="stylesheet">@font-face{width: calc(50ex - 0px)}</style>
+  </head>
+</html>
--- a/install/mod_pagespeed_test/css_minify_unicode_range_descriptor.html
+++ b/install/mod_pagespeed_test/css_minify_unicode_range_descriptor.html
@@ -1,0 +1,5 @@
+<html>
+  <head>
+     <style property="stylesheet">@font-face{unicode-range: U+0400-045F, U+0490-0491, U+04B0-04B1, U+2116;}</style>
+  </head>
+</html>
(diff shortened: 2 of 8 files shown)


oschaaf:
Just respun the tests, the CI test broke on issues caused by the recent migration of `modpagespeed.com`, which should be fixed now.

Changing the parser to support unicode-ranges would probably be a lot more work, right? Assuming the answer to that is 'yes', I think working on landing this specific fix makes sense.

jmarantz (approved):
Looks great; thanks for all the tests!

I'm not sure if going forward you need a new system-test for every functional corner-case.  It may be more maintainable over time to rely on unit-tests for those, and use system-tests to make sure that the integrated system is functional.

ashishk-1:
@oschaaf 
> Changing the parser to support unicode-ranges would probably be a lot more work, right? Assuming the answer to that is 'yes', I think working on landing this specific fix makes sense.

Yes, I went through codebase of third party css parser library, adding support to _unicode-range_ descriptor involves lot more work. For now, the fix will work correctly.

PR #1619, opened 2017-08-22
CSP: Handle restrictions on output URL, base

morlovich:
Just because we are permitted to read the input by CSP doesn't mean the output we produce 
(whether URL or inline) would be permitted by the policy.  As same inputs may occur on pages 
with different output policies, this introduces a render-time check, calling new hook PolicyPermitsRenderingon the RewriteContext at Render time, proceeding largely the same as if we were detached if it failed (modulo adding a comment). 

This change also extends the base-validity checking to be aware of CSP (logically a separate change, really..)

### DIFF
--- a/net/instaweb/rewriter/cache_extender.cc
+++ b/net/instaweb/rewriter/cache_extender.cc
@@ -65,13 +65,15 @@ const int64 kMinThresholdMs = Timer::kMonthMs;
 
 class CacheExtender::Context : public SingleRewriteContext {
  public:
-  Context(CacheExtender* extender, RewriteDriver* driver,
+  Context(RewriteDriver::InputRole input_role,
+          CacheExtender* extender, RewriteDriver* driver,
           RewriteContext* parent)
       : SingleRewriteContext(driver, parent,
                              NULL /* no resource context */),
-        extender_(extender) {}
+        input_role_(input_role), extender_(extender) {}
   virtual ~Context() {}
 
+  bool PolicyPermitsRendering() const override;
   virtual void Render();
   virtual void RewriteSingle(const ResourcePtr& input,
                              const OutputResourcePtr& output);
@@ -101,6 +103,7 @@ class CacheExtender::Context : public SingleRewriteContext {
   }
 
  private:
+  RewriteDriver::InputRole input_role_;
   CacheExtender* extender_;
   DISALLOW_COPY_AND_ASSIGN(Context);
 };
@@ -218,7 +221,8 @@ void CacheExtender::StartElementImpl(HtmlElement* element) {
 
       ResourceSlotPtr slot(driver()->GetSlot(
           input_resource, element, attributes[i].url));
-      Context* context = new Context(this, driver(), NULL /* not nested */);
+      Context* context = new Context(input_role, this, driver(),
+                                     NULL /* not nested */);
       context->AddSlot(slot);
       driver()->InitiateRewrite(context);
     }
@@ -237,7 +241,9 @@ void CacheExtender::StartElementImpl(HtmlElement* element) {
         if (slot == nullptr) {
           continue;
         }
-        Context* context = new Context(this, driver(), nullptr /* !nested */);
+        Context* context = new Context(
+              RewriteDriver::InputRole::kImg, this,
+              driver(), nullptr /* !nested */);
         context->AddSlot(RefCountedPtr<ResourceSlot>(slot));
         driver()->InitiateRewrite(context);
       }
@@ -258,6 +264,19 @@ void CacheExtender::Context::RewriteSingle(
   RewriteDone(
       extender_->RewriteLoadedResource(
           input_resource, output_resource, mutable_output_partition(0)), 0);
+}
+
+bool CacheExtender::Context::PolicyPermitsRendering() const {
+  if (num_output_partitions() == 1 && output(0).get() != nullptr
+      && output(0)->has_hash()) {
+    // This uses the InputRole rather than CspDirective variant to
+    // handle kUnknown (and to get bonus handling of kReconstruction,
+    // which wouldn't actually call this, but for which we still need to
+    // override).
+    return Driver()->IsLoadPermittedByCsp(
+        GoogleUrl(output(0)->url()), input_role_);
+  }
+  return true;  // e.g. failure cases -> still want to permit error to render.
 }
 
 void CacheExtender::Context::Render() {
@@ -398,12 +417,14 @@ RewriteResult CacheExtender::RewriteLoadedResource(
 }
 
 RewriteContext* CacheExtender::MakeRewriteContext() {
-  return new Context(this, driver(), NULL /*not nested*/);
+  return new Context(RewriteDriver::InputRole::kReconstruction, this,
+                     driver(), NULL /*not nested*/);
 }
 
 RewriteContext* CacheExtender::MakeNestedContext(
     RewriteContext* parent, const ResourceSlotPtr& slot) {
-  Context* context = new Context(this, NULL /* driver*/, parent);
+  Context* context = new Context(
+      RewriteDriver::InputRole::kUnknown, this, NULL /* driver*/, parent);
   context->AddSlot(slot);
   return context;
 }
--- a/net/instaweb/rewriter/cache_extender.cc
+++ b/net/instaweb/rewriter/cache_extender.cc
@@ -222,7 +222,7 @@ void CacheExtender::StartElementImpl(HtmlElement* element) {
       ResourceSlotPtr slot(driver()->GetSlot(
           input_resource, element, attributes[i].url));
       Context* context = new Context(input_role, this, driver(),
-                                     NULL /* not nested */);
+                                     nullptr /* not nested */);
       context->AddSlot(slot);
       driver()->InitiateRewrite(context);
     }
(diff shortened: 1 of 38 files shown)


net/instaweb/rewriter/cache_extender.cc
```diff
@@ -218,7 +221,8 @@ void CacheExtender::StartElementImpl(HtmlElement* element) {
 
       ResourceSlotPtr slot(driver()->GetSlot(
           input_resource, element, attributes[i].url));
-      Context* context = new Context(this, driver(), NULL /* not nested */);
+      Context* context = new Context(input_role, this, driver(),
+                                     NULL /* not nested */);
```
oschaaf: maybe change to nullptr? (there's a couple of candidates below that could also use that)

net/instaweb/rewriter/rewrite_context.cc
```diff
@@ -1996,13 +1998,28 @@ void RewriteContext::Harvest() {
 void RewriteContext::Render() {
 }
 
+bool RewriteContext::AreOutputsAllowedByCsp(CspDirective role) const {
+  if (Driver()->content_security_policy().empty()) {
+   return true;
+  }
+
+  for (const OutputResourcePtr& o : outputs_) {
+    if (o.get() != nullptr && o->has_hash() && o->has_url() &&
```
oschaaf, line 2007: Why is it necessary to consider `o->has_hash()` here?

oschaaf (commented):
Took a first pass at this, looks good so far. I'm leaving a request about s/NULL/nullptr and a question behind for now. I'll take another (closer) look later today.

net/instaweb/rewriter/rewrite_context.cc
morlovich: Hmm, it is a precondition for calling ->url(), though I don't remember off the top of my head when you would not have it.
Maybe I should run the tests with a DCHECK

(That is_url is different only for InlneOutputResource is kinda weird, too)

net/instaweb/rewriter/cache_extender_test.cc
```diff
@@ -1157,6 +1157,44 @@ TEST_F(CacheExtenderTest, VaryOrigin) {
                    StringPrintf(kCssFormat, cache_extended_css.c_str()));
 }
 
+TEST_F(CacheExtenderTest, RenderCsp) {
+  InitTest(100);
```
oschaaf: It's not clear to me what '100' means here (and in the lines below the same number shows up, should it maybe be constantized?)

oschaaf (approved):
One request on a magic number which isn't immediately clear, but this looks great to me otherwise. 

One question, perhaps out of scope for this change, but should the domain rewriting filter be made aware of CSP?

net/instaweb/rewriter/cache_extender_test.cc
morlovich: It's the TTL (in seconds) the fixture will give the resources. It mostly doesn't matter for this test (some of the others use AdvanceTime, etc), just has to be in a range that we would cache extend. Could probably use the existing 'kShortTtlSec ' constant, though.

oschaaf:
LGTM!

net/instaweb/rewriter/rewrite_context.cc
```diff
@@ -2016,10 +2033,24 @@ void RewriteContext::Propagate(bool render_slots) {
...
+                                      slot(0)->element());
+      }
+      else if (render_op == RenderOp::kRenderOnlyCspWarning) {
+        StringPiece name = id();
+        RewriteFilter* filter = Driver()->FindFilter(id());
+        if (filter != nullptr) {
+          name = filter->Name();
+        }
+        Driver()->InsertDebugComment(
+            StrCat("PageSpeed output (by ", name, ") not permitted by Content "
```
jmarantz, line 2048: I can't remember; do we need to escape debug comment strings?

net/instaweb/rewriter/scan_filter_test.cc
```diff
@@ -226,4 +226,44 @@ TEST_F(ScanFilterTest, CspParseOff) {
       GoogleUrl("http://www.example.org/foo.png"), CspDirective::kImgSrc));
 }
 
+TEST_F(ScanFilterTest, CspBase1) {
+  rewrite_driver()->AddFilters();
+  EnableDebug();
+  // The default base (the URL) is overridden by a base tag.
+  const char kTestName[] = "set_base";
```
jmarantz: static, here & below.

net/instaweb/rewriter/scan_filter_test.cc
```diff
@@ -226,4 +226,44 @@ TEST_F(ScanFilterTest, CspParseOff) {
...
+                           "</head>"));
+  EXPECT_FALSE(rewrite_driver()->other_base_problem());
+}
+
+
+TEST_F(ScanFilterTest, CspBase2) {
+  rewrite_driver()->AddFilters();
+  EnableDebug();
+  // The default base (the URL) is overridden by a base tag.
+  const char kTestName[] = "set_base";
```
jmarantz: static, here & below.

jmarantz (approved):

net/instaweb/rewriter/image_rewrite_filter_test.cc
```diff
@@ -4431,4 +4431,45 @@ TEST_F(ImageRewriteTest, BasicCsp) {
...
+TEST_F(ImageRewriteTest, RenderCsp) {
+  AddRecompressImageFilters();
+  rewrite_driver()->AddFilters();
+  EnableDebug();
+  AddFileToMockFetcher("images/a.jpg", kPuzzleJpgFile, kContentTypeJpeg, 100);
+  AddFileToMockFetcher("uploads/b.png", kPuzzleJpgFile, kContentTypeJpeg, 100);
+
+  // We want a policy that only affects output resources here (since the input
+  // one will apply at CreateInputResource time, before the cache lookup)
+  const char kCsp[] =
```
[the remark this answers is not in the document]
morlovich: Done

net/instaweb/rewriter/javascript_filter_test.cc
```diff
@@ -1457,6 +1486,29 @@ TEST_P(JavascriptFilterTest, InlineCsp) {
              "<!--Avoiding modifying inline script with CSP present-->"));
 }
 
+TEST_P(JavascriptFilterTest, CspBaseUri) {
+  InitFilters();
+  EnableDebug();
+  SetResponseWithDefaultHeaders(
+      "scripts/a.js", kContentTypeJavascript, kJsData, 100);
+
+  const char kCspAndBase[] =
```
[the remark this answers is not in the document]
morlovich: Done

net/instaweb/rewriter/js_inline_filter_test.cc
```diff
@@ -507,6 +507,31 @@ TEST_F(JsInlineFilterTest, NoFlushSplittingScriptTag) {
                output_buffer_);
 }
 
+TEST_F(JsInlineFilterTest, BasicCsp) {
+  SetHtmlMimetype();
+  AddFilter(RewriteOptions::kInlineJavascript);
+  EnableDebug();
+
+  const char kJs[] = "function id(x) { return x; }\n";
```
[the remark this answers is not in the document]
morlovich: Done x3

net/instaweb/rewriter/rewrite_context.cc
morlovich: InsertDebugComment does it.
https://github.com/pagespeed/mod_pagespeed/blob/04e3c5b42e2c7fdacb9dd0deaaeadbff2a3447d4/net/instaweb/rewriter/rewrite_driver.cc#L3235

net/instaweb/rewriter/scan_filter_test.cc
morlovich: Done x3

net/instaweb/rewriter/scan_filter_test.cc
morlovich: Done x3

net/instaweb/rewriter/cache_extender.cc
```diff
@@ -260,6 +266,19 @@ void CacheExtender::Context::RewriteSingle(
           input_resource, output_resource, mutable_output_partition(0)), 0);
 }
 
+bool CacheExtender::Context::PolicyPermitsRendering() const {
+  if (num_output_partitions() == 1 && output(0).get() != nullptr
+      && output(0)->has_hash()) {
```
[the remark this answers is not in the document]
morlovich: re: num_output_partitions with null resource: an optimization failed world looks like that.
(or an inline one, which isn't relevant here)

morlovich (commented):
(Not sure if I flushed my replies or not. Also looks like I'll need to rebase)

net/instaweb/rewriter/common_filter.cc
```diff
@@ -159,6 +160,17 @@ void CommonFilter::ResolveUrl(StringPiece input_url, GoogleUrl* out_url) {
   }
 }
 
+bool CommonFilter::IsRelativeUrlLoadPermittedByCsp(
+    StringPiece url, CspDirective role) {
+  GoogleUrl abs_url;
+  ResolveUrl(url, &abs_url);
+  if (abs_url.IsWebValid()) {
+    return driver()->IsLoadPermittedByCsp(abs_url, role);
+  } else {
```
jmarantz, line 169: Could remove the else.

net/instaweb/rewriter/rewrite_context.cc
jmarantz: Did your run the tests with a DCHECK?

jmarantz (approved):

morlovich:
re: DCHECK (why can't I reply to it): I did, and it didn't hit it.

net/instaweb/rewriter/common_filter.cc
morlovich: I like the symmetry...

PR #1626, opened 2017-09-04
Add cPanel EasyApache 4 RPM build support

kmark:
This is an alternative to pagespeed/cpanel which appears to now be
unmaintained.

The advantages for both server admins and maintainers, are:

 * The default pagespeed.conf is shared between the main mod_pagespeed
   targets (DEB and RPM) and cPanel users. Previously cPanel users
   would need to manually check the main repository to see if there were
   any relevant changes. Or repository maintainers would need to manually
   copy over the changes every time the default configuration changed.
 * cPanel (shared hosting) specific default security tweaks
 * Proper versioning of the cPanel package
 * Package dependencies and other package metadata also stay in sync
 * Ability to build beta releases
 * No need to maintain a separate repository (pagespeed/cpanel)
 * No need to compile or download a prebuilt mod_pagespeed RPM to then
   extract and shuffle around the contents to create an EasyApache
   compatible RPM.

### DIFF
--- a/build/linux_package_cpanel.gypi
+++ b/build/linux_package_cpanel.gypi
@@ -1,0 +1,34 @@
+# Copyright (c) 2010 The Chromium Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+{
+  'type': 'none',
+  'dependencies': [
+    'all.gyp:All',
+    'linux_installer_configs',
+  ],
+  'actions': [
+    {
+      'action_name': 'linux_package_deb_<(channel)_action',
+      'process_outputs_as_sources': 1,
+      'inputs': [
+        '<(cpanel_build)',
+        '<(PRODUCT_DIR)/install/rpm/mod-pagespeed.spec.template',
+        '<@(packaging_files_binaries)',
+        '<@(packaging_files_common)',
+        '<@(packaging_files_cpanel)',
+      ],
+      'outputs': [
+        '<(PRODUCT_DIR)/ea-apache24-mod_pagespeed-<(channel)-<(version)-r<(revision).<(rpm_arch).rpm',
+      ],
+      'action': [ '<@(rpm_cmd)', '-c', '<(channel)', '-p' ],
+    },
+  ],
+}
+
+# Local Variables:
+# tab-width:2
+# indent-tabs-mode:nil
+# End:
+# vim: set expandtab tabstop=2 shiftwidth=2:
--- a/install/rpm/build.sh
+++ b/install/rpm/build.sh
@@ -26,25 +26,31 @@ stage_install_rpm() {
   prep_staging_rpm
   stage_install_common
   echo "Staging RPM install files in '${STAGEDIR}'..."
-  process_template "${BUILDDIR}/install/common/rpmrepo.cron" \
-    "${STAGEDIR}/etc/cron.daily/${PACKAGE}"
-  chmod 755 "${STAGEDIR}/etc/cron.daily/${PACKAGE}"
+  if [ "$CPANEL" = false ]; then
+    process_template "${BUILDDIR}/install/common/rpmrepo.cron" \
+      "${STAGEDIR}/etc/cron.daily/${PACKAGE}"
+    chmod 755 "${STAGEDIR}/etc/cron.daily/${PACKAGE}"
+  fi
 
   # For CentOS, the load and conf files are combined into a single
   # 'conf' file. So we install the load template as the conf file, and
   # then concatenate the actual conf file.
   process_template "${BUILDDIR}/install/common/pagespeed.load.template" \
-    "${STAGEDIR}${APACHE_CONFDIR}/pagespeed.conf"
+    "${STAGEDIR}${APACHE_CONFDIR}/${CPANEL_CONF_PREFIX}pagespeed.conf"
   process_template "${BUILDDIR}/install/common/pagespeed.conf.template" \
     "${BUILDDIR}/install/common/pagespeed.conf"
   cat "${BUILDDIR}/install/common/pagespeed.conf" >> \
-    "${STAGEDIR}${APACHE_CONFDIR}/pagespeed.conf"
+    "${STAGEDIR}${APACHE_CONFDIR}/${CPANEL_CONF_PREFIX}pagespeed.conf"
+  if [ "$CPANEL" = true ]; then
+    cat "${BUILDDIR}/install/rpm/pagespeed.cpanel.conf" >> \
+      "${STAGEDIR}${APACHE_CONFDIR}/${CPANEL_CONF_PREFIX}pagespeed.conf"
+  fi
   install -m 755 "${BUILDDIR}/js_minify" \
     "${STAGEDIR}/usr/bin/pagespeed_js_minify"
-  chmod 644 "${STAGEDIR}${APACHE_CONFDIR}/pagespeed.conf"
+  chmod 644 "${STAGEDIR}${APACHE_CONFDIR}/${CPANEL_CONF_PREFIX}pagespeed.conf"
   install -m 644 \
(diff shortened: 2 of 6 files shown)


kmark:
Travis build seemed to fail due to a network error. If anyone can restart the build I'd appreciate it.

oschaaf:
@kmark restarted!

kmark:
@oschaaf Thanks!

build/linux_package_cpanel.gypi
```diff
@@ -0,0 +1,34 @@
...
+      ],
+      'outputs': [
+        '<(PRODUCT_DIR)/ea-apache24-mod_pagespeed-<(channel)-<(version)-r<(revision).<(rpm_arch).rpm',
+      ],
+      'action': [ '<@(rpm_cmd)', '-c', '<(channel)', '-p' ],
+    },
+  ],
+}
+
+# Local Variables:
```
oschaaf, line 30: There's no formatting hints included in any sourcecode of this repo, so this would be an exception.
Could you remove this? (If we'd include this, I'd prefer it as a project-wide change / guideline. But then it would make sense to first discuss this in a public thread so everyone has a chance to share their opinion/thoughts. So for now, leaving it out would probably  be easiest).

install/rpm/build.sh
```diff
@@ -135,7 +147,7 @@ verify_channel() {
 }
 
 process_opts() {
-  while getopts ":o:b:c:a:h" OPTNAME
+  while getopts ":o:b:c:a:ph" OPTNAME
```
oschaaf: Does this accidentally drop `-h`?

build/linux_package_cpanel.gypi
oschaaf: Generally, we don't do this (I don't think any file in the repo includes vim formatting directives). Could you leave this out?

install/rpm/build.sh
oschaaf, line 150: Does this accidentally drop the `-h` option?

oschaaf (changes requested):
Thanks for this contribution! I left in a few small remarks.

I'd love to hear from other contributors what they think about including this and possibly `deprecating pagespeed/cpanel` (I have yet to familiarize myself with that repo). For now I have no objections but I would like to test this change - if the concensus is to merge this. Could you add description on how we can do that?

Going forward, if we merge this it would be good if we can keep this working. Would it be hard to design a test for that? Would you be willing to look into that?
If we merge this, some documentation may also be helpful I think. For example we could add a page on modpagespeed.com with contents similar to the readme over at https://github.com/pagespeed/cpanel to maximize discoverability?

oschaaf:
(My apologies, some review comments somehow ended up duplicated in the comments, I'm not sure what happened. I deleted the duplicates)

build/linux_package_cpanel.gypi
kmark: Sure. It's worth noting that this file was based off [build/linux_package_rpm.gypi](https://github.com/pagespeed/mod_pagespeed/blob/93541308744ebc586578e531cb54072b0ddb1eda/build/linux_package_rpm.gypi#L30) which includes the formatting hints.

install/rpm/build.sh
kmark: It shouldn't, no. If my understanding is correct options without arguments do not have a colon after them. A quick test shows `-h` still working.

kmark:
> For now I have no objections but I would like to test this change - if the concensus is to merge this. Could you add description on how we can do that?

I'd have to familiarize myself with the testing system in use. If we're currently testing the RPM package I would think we could end up duplicating most of those tests to work with the cPanel specific RPM. If we're not testing any packages right now it might take a little more time/effort.

> If we merge this, some documentation may also be helpful I think.

I agree. I was considering submitting a documentation patch since the current docs do not reflect how to build the current packages. For instance, you must use the `linux_package_rpm_stable` target not `linux_package_rpm`. The nice thing about this patch as opposed to the cpanel repository is that the instructions for building the cPanel RPM are exactly the same as the normal RPM. Just a different target. We'd only need to mention in the docs the recommended way to install the RPM (yum localinstall) and remove it (also yum) and potentially where the config files are located. When you install/remove an EasyApache RPM through yum cPanel has hooks that automatically restart Apache for you.

build/linux_package_cpanel.gypi
oschaaf: I see, looks good to me then. (Perhaps we should clean up the inconsistency at some point in the future, and either include this in all .gypi files, or axe it out everywhere).

install/rpm/build.sh
oschaaf: I see, great!

oschaaf (approved):
LGTM, no more code level comments from me. 

Re: testing

Tracing what happens from here on may be a good starting point, if we can run something in Travis CI that would be great imho: https://github.com/pagespeed/mod_pagespeed/blob/master/install/build_release.sh#L119

Nirjonadda:
As I know cPanel EasyApache 4 support for mod_pagespeed ?

Here: https://features.cpanel.net/topic/add-support-for-mod_pagespeed-to-easyapache4

kmark:
Hi @Nirjonadda. cPanel has support for mod_pagespeed insofar as it's in their experimental repository. However from my understanding there are no current plans for it to be moved to the stable repo and by virtue of it being in the experimental repo the actual support for it is probably little to none. I am not sure if the module will be updated either for security patches. Even enabling the experimental repo is probably not a good idea in production.

Cumulatively these can be show stoppers to using mod_pagespeed in production service with cPanel. This patch certainly makes it viable. And because it's part of the same build process as the normal RPM and DEB there's very little preventing official binaries from Google. This patch could also pave the way for official downstream support in cPanel as the bulk of the work, I would assume, is done for them.

It may also be worth noting that the experimental cPanel build appears to use the pagespeed/cpanel repo which has an insecure default configuration because it does not restrict access to the administrative features. It also allows the viewing of statistics "only from localhost" which is not secure for shared hosting services which allow their clients to execute code which can potentially read from localhost.

install/rpm/build.sh
```diff
@@ -26,25 +26,31 @@ stage_install_rpm() {
...
+      "${STAGEDIR}/etc/cron.daily/${PACKAGE}"
+    chmod 755 "${STAGEDIR}/etc/cron.daily/${PACKAGE}"
+  fi
 
   # For CentOS, the load and conf files are combined into a single
   # 'conf' file. So we install the load template as the conf file, and
   # then concatenate the actual conf file.
   process_template "${BUILDDIR}/install/common/pagespeed.load.template" \
-    "${STAGEDIR}${APACHE_CONFDIR}/pagespeed.conf"
+    "${STAGEDIR}${APACHE_CONFDIR}/${CPANEL_CONF_PREFIX}pagespeed.conf"
```
keesspoelstra: Shouldn't we rename CPANEL_CONF_PREFIX into something like PAGESPEED_CONF_PREFIX
It does not sit well with me to have very specifically named vars in the global flow.

keesspoelstra (changes requested):

install/rpm/build.sh
kmark: Agreed, considering there are other circumstances in which environments other than cPanel prefer a prefix.

oschaaf:
Anyone else interested in reviewing this? Any thoughts?

install/rpm/build.sh
kmark: This has been addressed in the latest commit

kmark:
Just commenting to confirm that all requested changes have been addressed.

oschaaf:
@keesspoelstra your changes have been processed, lgty? @jmarantz any thoughts on this?

jmarantz (approved):

oschaaf:
@kmark Merged -- thanks! 
Would you be willing to help out with the next steps? (Update the docs, deprecate the pagespeed/cpanel repo)

PR #1627, opened 2017-09-08
Added support for redis database selection(Fixes #1609)

### LINKED ISSUE #1609: Defining Redis DBs
Hi there,

I am wondering if there is a way to specify which Redis DB to use when adding Redis cache. 

I couldn't find any documentation on it, normally you would set some kind of "Define_redis_DB = #" statement. I use Redis for many things so I am needing to set pagespeed to a specific DB ID (16)

Thanks

ashishk-1:
"ModPagespeedRedisDatabaseSelectIndex index" property added in conf file.
When enabled, selects the redis database mentioned by index.

- After redis database connection, executing a redis command to select database. 
- In redis_cache_tests, select command is added in HandleClientConnection before get request. This tests the mock redis socket connection and get request.
- Default redis database index is 0.

### DIFF
--- a/html/doc/system.html
+++ b/html/doc/system.html
@@ -483,9 +483,9 @@ total_items:           4166384
     </p>
 <dl>
   <dt>Apache:<dd><pre class="prettyprint"
-     >ModPagespeedRedisDatabaseSelectIndex index</pre>
+     >ModPagespeedRedisDatabaseIndex index</pre>
   <dt>Nginx:<dd><pre class="prettyprint"
-     >pagespeed RedisDatabaseSelectIndex index;</pre>
+     >pagespeed RedisDatabaseIndex index;</pre>
 </dl>
 
     <h2 id="flush_cache">Flushing PageSpeed Server-Side Cache</h2>
--- a/html/doc/system.html
+++ b/html/doc/system.html
@@ -475,6 +475,18 @@ total_items:           4166384
   <dt>Nginx:<dd><pre class="prettyprint"
      >pagespeed RedisReconnectionDelayMs timeout_in_milliseconds;</pre>
 </dl>
+    <p>
+      You can also select the Redis logical database having the specified numeric index.
+      This setting defaults to database index 0.
+      Note that the Redis clustering feature does not allow specifying a database index,
+      and therefore is incompatible with this setting:
+    </p>
+<dl>
+  <dt>Apache:<dd><pre class="prettyprint"
+     >ModPagespeedRedisDatabaseSelectIndex index</pre>
+  <dt>Nginx:<dd><pre class="prettyprint"
+     >pagespeed RedisDatabaseSelectIndex index;</pre>
+</dl>
 
     <h2 id="flush_cache">Flushing PageSpeed Server-Side Cache</h2>
     <p>
--- a/pagespeed/system/redis_cache.cc
+++ b/pagespeed/system/redis_cache.cc
@@ -78,6 +78,7 @@ namespace net_instaweb {
 // TODO(yeputons): consider removing limit on amount of redirections.
 static const int kMaxRedirections = 1;
 static const int kDefaultDatabaseIndex = 0;
+static const int kRedisDatabaseIndexNotSet = -1;
 
 const char kRedisClusterRedirections[] = "redis_cluster_redirections";
 const char kRedisClusterSlotsFetches[] = "redis_cluster_slots_fetches";
@@ -621,7 +622,7 @@ bool RedisCache::Connection::EnsureConnection() {
 
 bool RedisCache::Connection::EnsureDatabaseSelection() {
   // dont select database if database index property not specified in config
-  if (database_index_ != -1) {
+  if (database_index_ != kRedisDatabaseIndexNotSet) {
     RedisReply reply = RedisCommand(StrCat("SELECT ",
             IntegerToString(database_index_)).c_str(), REDIS_REPLY_STRING);
     if (reply == nullptr) {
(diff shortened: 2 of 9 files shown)


html/doc/system.html
```diff
@@ -475,6 +475,18 @@ <h4 id="redis">Configuring Redis</h4>
...
 </dl>
+    <p>
+      You can also select the Redis logical database having the specified numeric index.
+      This setting defaults to database index 0.
+      Note that the Redis clustering feature does not allow specifying a database index,
+      and therefore is incompatible with this setting:
+    </p>
+<dl>
+  <dt>Apache:<dd><pre class="prettyprint"
+     >ModPagespeedRedisDatabaseSelectIndex index</pre>
```
jmarantz: Is this a DatabaseSelectIndex, or are we selecting a DatabaseIndex?  Just want to make sure we get a name here that makes sense to Redis people.  If it's really called a DatabaseSelectIndex then this is the right spelling, and we should also change the param below from "int database_index" to "int database_select_index". 

  If we are selecting a DatabaseIndex then we should remove the "Select" from the name.

pagespeed/system/redis_cache.cc
```diff
@@ -599,6 +619,18 @@ bool RedisCache::Connection::EnsureConnection() {
   return state_ == kConnected;
 }
 
+bool RedisCache::Connection::EnsureDatabaseSelection() {
+  RedisReply reply = RedisCommand(StrCat("SELECT ",
```
jmarantz: If this is incompatible with clustering, then maybe we should conditionalize this on options->has_database_index().

Even if it's not incompatible with clustering, should we just omit the explicit db index selection if not specified in the config?  I'm not enough of a redis expert to know the right thing here, but if it's working today maybe we shouldn't bother sending the command unless explicitly configured.

pagespeed/system/redis_cache_test.cc
```diff
@@ -45,6 +45,7 @@ namespace net_instaweb {
 namespace {
   static const int kReconnectionDelayMs = 10;
   static const int kTimeoutUs = 100 * Timer::kMsUs;
+  static const int kDatabaseIndex = 0;
```
jmarantz: It's fine to have a constant index for the indexing tests.  But can you add a test that opens 2 clients with the same redis server, different DB indexes.  Make sure a given key in one client is independent of the same key in the other one.

pagespeed/system/system_rewrite_options.h
```diff
@@ -187,6 +188,9 @@ class SystemRewriteOptions : public RewriteOptions {
   int64 redis_timeout_us() const {
     return redis_timeout_us_.value();
   }
+  int redis_database_index() const {
+    return redis_database_index_.value();
+  }
```
jmarantz, line 193: add has_redis_database_index() modeled after the other has_ methods.

oschaaf:
Travis ran into one failing test, which looks relevant to this change:
```
[ RUN      ] RedisCacheTest.ShutDownDuringConnection
pagespeed/system/redis_cache_test.cc:433: Failure
Value of: recvSize
  Actual: 0
Expected: ((sizeof(ArraySizeHelper(kSelectRequest))) - 1)
Which is: 23
pagespeed/system/redis_cache_test.cc:435: Failure
Value of: requestBuf
  Actual: ""
Expected: kSelectRequest
Which is: "*2\r\n$6\r\nSELECT\r\n$1\r\n0\r\n"
[  FAILED  ] RedisCacheTest.ShutDownDuringConnection (1 ms)
```

html/doc/system.html
ashishk-1: We are selecing database index. So modifying property name to ModPagespeedRedisDatabaseIndex

pagespeed/system/redis_cache.cc
ashishk-1: Added method has_redis_database_index() to check property from config. Selecting database only if property present in config.

ashishk-1:
Added patch with the following changes:

- property name changed to ModPagespeedRedisDatabaseIndex
- added unit test for multiple databases, also rearrangement of unit test code
- selecting database only when property specified in conf file
- documentation updated

pagespeed/system/redis_cache_test.cc
```diff
@@ -73,35 +75,40 @@ class RedisCacheTest : public CacheTestBase {
...
 
+  void InitRedisCache(int database_index = 0) {
+    cache_.emplace_back(new RedisCache("localhost", redis_port_env_,
+                            thread_system_.get(), &handler_, &timer_,
+                            kReconnectionDelayMs, kTimeoutUs, &statistics_,
+                            database_index));
+    cache_.back()->StartUp();
+  }
+
   void InitRedisWithCustomServer() {
```
oschaaf, line 98: InitRedisWithCustomDatabaseIndex may be a better name

pagespeed/system/redis_cache.cc
```diff
@@ -599,6 +619,21 @@ bool RedisCache::Connection::EnsureConnection() {
   return state_ == kConnected;
 }
 
+bool RedisCache::Connection::EnsureDatabaseSelection() {
+  // dont select database if database index property not specified in config
+  if (database_index_ != -1) {
```
oschaaf: Maybe constantize -1 to `kRedisDatabaseIndexUnset`?

oschaaf (commented):
Two small requests related to naming, those aside this looks good to me. (Also: Travis passed, great!)

ashishk-1:
Added review changes:
- using kRedisDatabaseIndexNotSet as -1 in redis_cache when property not mentioned in config.
- InitRedisCache method renamed to InitRedisWithCustomDatabaseIndex.

oschaaf (approved):
lgtm!

ashishk-1:
Removed default argument.

jmarantz (approved):
Thanks!

PR #1629, opened 2017-09-12
Travis network errors: use github mirrors for git.apache.org

oschaaf:


.travis.yml
```diff
@@ -15,8 +15,10 @@ before_install:
   - sudo add-apt-repository -y ppa:git-core/ppa
   - sudo apt-get update -q
   - sudo apt-get install -q -y git
+  # git.apache.org is the canonical repo but throttles network connections.
+  # using github.com seems to reduce network errors a lot.
```
jmarantz, line 19: Seems OK to me but I wonder if someone will get upset at some point.

jmarantz (approved):

.travis.yml
oschaaf: @jmarantz is your concern that we would be using too many concurrent jobs for fetching submodules hosted on github.com?: I'm not sure if this change will increase concurrency (much) towards github.com: we're already using it for a couple of dependencies. We will be fetching more repositories though, but with the same amount of parallellism afaict.

One alternative is that we look into some kind of retry mechanism, on top of the git client's own retry mechanism, to catch timeouts and connection refused messages. We would have to wait for a couple of minutes before retrying.. But that may cause us to run into another issue more frequently: we're fairly close to the restriction Travis imposes on the test duration.

jmarantz: I think we should do this, and get prepared for Apache to ask whether we can do this without DOSing them at some point :)

I wonder if we can control the overall parallelism a bit to reduce load (and perhaps get the standard mechanism to work).  But let's make travis work this way and then iterate.

PR #1630, opened 2017-09-13
System tests: stabilize css rewriting edge cases tests

oschaaf:
These tests sometimes flaked in the more extensive checkin tests

### DIFF
--- a/pagespeed/system/system_tests/css_minify_calc_function_value_zero.sh
+++ b/pagespeed/system/system_tests/css_minify_calc_function_value_zero.sh
@@ -18,8 +18,6 @@
 start_test CSS minify with calc function and value 0.
 
 URL="$TEST_ROOT/css_minify_calc_function_value_zero.html?PageSpeedFilters=rewrite_css"
-RESPONSE_OUT=$(http_proxy=$SECONDARY_HOSTNAME $WGET_DUMP --header 'X-PSA-Blocking-Rewrite: psatest' $URL)
-
 # checking for minified css with unit retained for 0 value 
-MATCHES=$(echo "$RESPONSE_OUT" | fgrep -c "width:calc(50ex - 0px)")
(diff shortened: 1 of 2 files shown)


pagespeed/system/system_tests/css_minify_calc_function_value_zero.sh
```diff
@@ -17,8 +17,8 @@
 
 start_test CSS minify with calc function and value 0.
 
-URL="$TEST_ROOT/css_minify_calc_function_value_zero.html?PageSpeedFilters=+inline_css"
-RESPONSE_OUT=$(http_proxy=$SECONDARY_HOSTNAME $WGET_DUMP $URL)
+URL="$TEST_ROOT/css_minify_calc_function_value_zero.html?PageSpeedFilters=rewrite_css"
+RESPONSE_OUT=$(http_proxy=$SECONDARY_HOSTNAME $WGET_DUMP --header 'X-PSA-Blocking-Rewrite: psatest' $URL)
```
jmarantz: I had traditionally been avoiding using X-PSA-Blocking-Rewrite for de-flaking positive tests.  The main reason it's there is for negative tests, e.g. I expect a disallowed resource to never be optimized.

The reason I don't want to use that for positive tests is that it deviates from the production path.  And the mechanism we use to avoid the flakes in most cases is fetch_until.  Is there some reason that won't work here?

oschaaf: @jmarantz `fetch_until` should work, and there's nothing that speaks for sticking with `X-PSA-Blocking-Rewrite` in this change. I'll amend this to use `fetch_until`

ashishk-1 (approved):

oschaaf:
@jmarantz are you happy with the way this looks now?

jmarantz (approved):
Thanks Otto, looks perfect.

PR #1641, opened 2017-10-10
Fix unexpected close-tag 'br' in server_context.cc

dev:
Use <br> instead of </br>.
Pagespeed noticed this:
[Tue, 10 Oct 2017 07:01:23 GMT] [Info] [26724] [http://contimacpagespeed.dmtest.be/pagespeed_admin/cache:182] Unexpected close-tag `br', no tags are open

### DIFF
--- a/net/instaweb/rewriter/server_context.cc
+++ b/net/instaweb/rewriter/server_context.cc
@@ -1357,7 +1357,7 @@ GoogleString ServerContext::ShowCacheForm(StringPiece user_agent) {
       "  URL: <input id=metadata_text type=text name=url size=110 /><br>\n"
       "  User-Agent: <input id=user_agent type=text size=103 name=user_agent ",
       ua_default,
-      "/></br> \n",
+      "/><br> \n",
       "  <input id=metadata_submit type=submit "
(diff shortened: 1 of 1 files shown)


dev:
I signed it!

net/instaweb/rewriter/server_context.cc
```diff
@@ -1357,7 +1357,7 @@ GoogleString ServerContext::ShowCacheForm(StringPiece user_agent) {
       "  URL: <input id=metadata_text type=text name=url size=110 /><br>\n"
       "  User-Agent: <input id=user_agent type=text size=103 name=user_agent ",
       ua_default,
-      "/></br> \n",
+      "/><br> \n",
```
oschaaf, line 1360: I think the right thing to do here would be to move the slash so this ends up like `<br/>`?

dev: The `/` character is optional for `<br>` (`<br/>`). But if you like `<br/>` more than `<br>` then we should do that! :)

- https://stackoverflow.com/a/20833968

dev: :smiley:

PR #1646, opened 2017-10-13
Don't special-case sending webp to PSI

jmarantz:
Remove special-case serving of webp to PSI (which lacks accept:image:webp) due to difficulties transmitting user-agent through cloudfront.  See https://github.com/pagespeed/mod_pagespeed/issues/1585

Note that webp may still be sent to PSI, depending on what's in the user-agent, and how AllowVary is set up.  But this is like any other pre-accept-header chrome.  What was broken before was ipro-rewriting to webp without the accept:image/webp header.


oschaaf (approved):
No comments, LGTM

PR #1647, opened 2017-10-14
change some flaky system-tests to use fetch_all, to help them pass valgrind

jmarantz:


oschaaf (approved):
So to summarize: when we use `fetch_until`, we should combine it `fgrep` and add the `-save` flag, is that correct?

LGTM

oschaaf:
(The Travis job timed out -- I just restarted it)

jmarantz:
Actually now that I think of it, the -save flag is not necessary in either of these new cases.  I will remove them.  The -save flag is needed if, after the fetch_until, you want to do more analysis on the output.

ashishk-1 (approved):
LGTM

oschaaf:
@jmarantz is this ready to go in?

PR #1667, opened 2017-11-01
Csp unit tests: default honor-csp to on

oschaaf:


jmarantz (commented):
What do you think about moving this to the Setup() method in the cap tests?

oschaaf:
@jmarantz The thing is that doing at that level woul be a change that hits a lot of files, what I was trying to avoid:
https://github.com/pagespeed/mod_pagespeed/search?utf8=%E2%9C%93&q=Csp&type=

Having said that: I would be up for it if you see a clear upside to doing it at that level?

jmarantz (approved):
I don't think it's *that* many files. I think it's:

csp_test.cc
javascript_filter_test.cc
scan_filter_test.cc
image_rewrite_filter_test.cc
js_combine_filter_test.cc
css_filter_test.cc

I figured that out with:

cd mod_pagespeed
find . -name "*_test.cc" -exec grep -n Csp {} /dev/null \;

I couldn't figure out how to get that "look for substring in files ending with _test.cc" with github search.

However given you are trying to get a release out, I think this change makes sense as you have it.

PR #1679, opened 2017-11-08
Add a missing link target, and add a missing space.

jeffkaufman:


oschaaf:
Thanks for catching these things, pushing them out to modpagespeed.com right away

PR #1691, opened 2017-12-06
RerwriteDriver: Clear srcset_collection when failing InitiateRewrite

oschaaf:
Fix for https://github.com/pagespeed/ngx_pagespeed/issues/1514

### DIFF
--- a/net/instaweb/rewriter/rewrite_driver.cc
+++ b/net/instaweb/rewriter/rewrite_driver.cc
@@ -2873,6 +2873,7 @@ bool RewriteDriver::InitiateRewrite(RewriteContext* rewrite_context) {
   // us with a partially constructed slot-graph.
   if (!can_rewrite_resources_) {
     if (rewrites_.empty()) {
+      srcset_collections_.clear();
       rewrite_context->DetachSlots();
       delete rewrite_context;
       return false;


net/instaweb/rewriter/rewrite_driver.cc
```diff
@@ -2873,6 +2873,7 @@ bool RewriteDriver::InitiateRewrite(RewriteContext* rewrite_context) {
   // us with a partially constructed slot-graph.
   if (!can_rewrite_resources_) {
     if (rewrites_.empty()) {
+      srcset_collections_.clear();
```
jmarantz, line 2876: first of all, great catch.  I think Maks needs to weigh in on if this is the right thing.  If it is, I wonder if we should factor out a ClearSlots() helper that has:

    slots_.clear();
    inline_slots_.clear();
    inline_attribute_slots_.clear();
    srcset_collections_.clear();

and call that both here and on line 816.

jmarantz (commented):
Also, do you have a way of capturing the stimulus that caused the crash as a unit test?

net/instaweb/rewriter/rewrite_driver.cc
oschaaf: I agree that would make sense .. so if Maks blesses this change I'll factor out `RewriteDriver::ClearSlots()`

oschaaf:
My current approach to reproduce isn't good to go as a system test. But now that we know where to look, it may be possible to cover this in a c++ test.. I'll have a look.

oschaaf:
Hmm a c++ test may not be that easy. 

After condensing things in the repro I currently have, I think it's actually simple, we need these ingredients:
- a vhost with a memcached configuration that points towards an address where no memcached server is listening
- a page with the following html should suffice:

```html
<!DOCTYPE html>
<html>
<body>
<img src=1.png  srcset='1.png? 1.5x, 1.png 2x,1.png'/>
</body>
</html>
````

-  Perform two requests on the test page. The second reply should be a 200/OK with presence of `</html>` in the entity body.

I'll add that.

oschaaf:
Following a discussion per email, this PR just squelches a symptom. Closing.

PR #1697, opened 2017-12-14
Add CspPolicy::ToIdentityString(). Use it in JS Combiner cache key suffix

oschaaf:
This change affects the way we handle dynamic policies (as well as occasional policy updates).
This PR is to review the general approach. To complete this update an audit of all filters is probably needed to see which filter cache key suffixes need changes.

(The redirecting fetcher PR contains some end-to-end tests on the way the partitioning behaves when different policies are in play for the same html url, I'm splitting out a series of smaller PR's to unbloat that one so review becomes more manageable)

### DIFF
--- a/net/instaweb/rewriter/csp_test.cc
+++ b/net/instaweb/rewriter/csp_test.cc
@@ -922,6 +922,18 @@ TEST(CspContext, BaseUri) {
   }
 }
 
+TEST(CspIdentityStringTest, Quoted) {
+  CspContext ctx;
+  ctx.AddPolicy(CspPolicy::Parse("default-src https:; img-src *"));
+  ctx.AddPolicy(CspPolicy::Parse(""));
+  ctx.AddPolicy(CspPolicy::Parse("style-src *.example.com"));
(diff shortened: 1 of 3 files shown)


morlovich:
So, high-level question:  
If this hash function always produces different outputs for distinct policies, there is no need for a render-time check; however verifying that it may be tricky, as I would have to reason about what happens when something has one of those characters you used in joins inside, the differences behind empty vs. missing rules, and all that. 

If there is a check, then we can skp over that analysis -except- you may get some non-rewriting flakiness 
if it's not quite 1-to-1. Probably still preferable given this is a security property..

(We would probably want to use it for other filters to, though; it may also make sense to summarize only pairs like style-src + default-src, since other things are irrelevant).

oschaaf:
I'll take a closer look at empty vs missing, and also at the delimiters used while constructing the identity string for hashing, thanks for pointing these things out.

Also you're right: it should be possible to optimize this by only considering pairs we care about when computing the cache key. combine_js would only care about (potentially) script-related directives, I think? 

Currently this change only has wired up `js_combine_filter.cc` to materialize something concrete.. when we're happy with the way this works and looks, I'll look into making changes to other filters.

Lastly, +1 to continue render time checking, absolutely.

Issue #1514, opened 2017-03-14
Inline JavaScript produces invalid HTML


Dreamsorcerer:
If a <script> tag with the defer attribute is inlined, it leaves the defer attribute in the script tag, but this then fails HTML validation:

> Element script must not have attribute defer unless attribute src is also specified.

When inlining JS, it should probably remove the defer attribute from the tag.

jmarantz:
Actually I suspect the right remedy for this is to avoid inlining scripts that are marked as 'defer'.  Note from https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script.  In particular, order needs to be preserved between multiple scripts with 'defer' attributes, and inlining some of them would surely violate that expectation.

This Boolean attribute is set to indicate to a browser that the script is meant to be executed after the document has been parsed, but before firing DOMContentLoaded.
This attribute must not be used if the src attribute is absent (i.e. for inline scripts), in this case it would have no effect.

To achieve a similar effect for dynamically inserted scripts use async=false instead. Scripts with the defer attribute will **execute in the order in which they appear** in the document.

PR #1698, opened 2017-12-20
Fix for nps issue #1514

oschaaf:
Regression test plus fix for https://github.com/pagespeed/ngx_pagespeed/issues/1514
(@morlovich: your comments from before this PR have been addressed)

### DIFF
--- a/net/instaweb/rewriter/public/srcset_slot.h
+++ b/net/instaweb/rewriter/public/srcset_slot.h
@@ -52,7 +52,7 @@ class SrcSetSlotCollection : public RefCounted<SrcSetSlotCollection> {
 
     // The slot owns us. Note that some of these may be nullptr if a resource
     // couldn't be created.
-    SrcSetSlot* slot;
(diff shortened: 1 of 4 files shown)


net/instaweb/rewriter/public/srcset_slot.h
```diff
@@ -50,9 +50,11 @@ class SrcSetSlotCollection : public RefCounted<SrcSetSlotCollection> {
     GoogleString url;
     GoogleString descriptor;
 
-    // The slot owns us. Note that some of these may be nullptr if a resource
+    // These reference are released by SrcSetSlotCollection::Detach(),
+    // which is called in RewriteDriver::FlushAsyncDone.
+    // Note that some of these may not be set if a resource
```
morlovich: Probably still worth elaborating that that breaks a reference-count cycle.

morlovich (approved):

oschaaf:
@morlovich Thanks -- updated the doc comment to mention the reference count cycle & merged

PR #1703, opened 2018-01-01
Handle Apache >2.4 mod_authz_host

eldk:
Hello,

According to this : https://httpd.apache.org/docs/2.4/en/mod/mod_access_compat.html, removes old directives and use new ones as explain here : https://httpd.apache.org/docs/2.4/en/upgrading.html#page-header .

Maybe you should keep : 
#Require forward-dns user.domain.tld
as example, it's to use domain verification only, but not reverse resolution (http://httpd.apache.org/docs/current/en/mod/mod_authz_host.html). This way a mobile user should use some dynamic ip's domains services easyly and admin should grant access to this user only .

Thanks,

Eric

### DIFF
--- a/install/common/pagespeed.conf.template
+++ b/install/common/pagespeed.conf.template
@@ -332,9 +332,9 @@
             </IfVersion>
         <IfVersion >= 2.4>
             Require local
-            #see https://httpd.apache.org/docs/current/fr/mod/mod_authz_host.html - Require all granted,
-            #Require ip, Require host or Require forward-dns,
-            #if access must be granted to other than local.
+            # see https://httpd.apache.org/docs/current/en/mod/mod_authz_host.html - Require all granted,
+            # Require ip, Require host or Require forward-dns,
+            # if access must be granted to other than local.
             </IfVersion>
         SetHandler pagespeed_admin
     </Location>
--- a/install/common/pagespeed.conf.template
+++ b/install/common/pagespeed.conf.template
@@ -325,11 +325,11 @@
       LoadModule version_module /usr/lib/httpd/modules/mod_version.so
     </IfModule>
     <Location /pagespeed_admin>
-        <IfVersion <2.4>
+        <IfVersion < 2.4>
             Order allow,deny
             Allow from localhost
             Allow from 127.0.0.1
-            <IfVersion>
+            </IfVersion>
         <IfVersion >= 2.4>
             Require local
             #see https://httpd.apache.org/docs/current/fr/mod/mod_authz_host.html - Require all granted,
@@ -339,11 +339,11 @@
         SetHandler pagespeed_admin
     </Location>
     <Location /pagespeed_global_admin>
-        <IfVersion <2.4>
+        <IfVersion < 2.4>
             Order allow,deny
             Allow from localhost
             Allow from 127.0.0.1
-            <IfVersion>
+            </IfVersion>
         <IfVersion >= 2.4>
             Require local            
             </IfVersion>
--- a/install/common/pagespeed.conf.template
+++ b/install/common/pagespeed.conf.template
@@ -321,15 +321,19 @@
     # and change server state, such as statistics, caches, and
     # messages.  This might be appropriate in an experimental setup.
     <Location /pagespeed_admin>
-        Order allow,deny
-        Allow from localhost
-        Allow from 127.0.0.1
+        Require all granted
+        Require local
+        #Require ip myremoteip
+        #Require host mydomain.tld
+        #Require forward-dns user.domain.tld
         SetHandler pagespeed_admin
     </Location>
     <Location /pagespeed_global_admin>
-        Order allow,deny
-        Allow from localhost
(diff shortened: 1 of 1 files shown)


install/common/pagespeed.conf.template
```diff
@@ -321,15 +321,19 @@
     # and change server state, such as statistics, caches, and
     # messages.  This might be appropriate in an experimental setup.
     <Location /pagespeed_admin>
-        Order allow,deny
-        Allow from localhost
-        Allow from 127.0.0.1
+        Require all granted
+        Require local
```
oschaaf: Are these two 'Require' lines equivalent to the above three that you removed?

install/common/pagespeed.conf.template
```diff
@@ -321,15 +321,19 @@
...
+        #Require ip myremoteip
```
oschaaf: Do we need these commented lines? If so, can you add a comment as to document why we need them? :-)

install/common/pagespeed.conf.template
jmarantz: can you put these under ifdef so this still works on 2.2?  Or will it work on 2.2 as is?

Or are we deciding to remove support for 2.2?

oschaaf: Apache 2.2 still looks pretty relevant, i would vote for continued support?:
https://trends.builtwith.com/Web-Server/Apache-2.2

eldk: Hello,
@oschaaf @jmarantz 
>Are these two 'Require' lines equivalent to the above three that you removed?

yes, they are.

eldk: @jmarantz 
>can you put these under ifdef so this still works on 2.2? Or will it work on 2.2 as is?

I will put ifdef, as it can't work with apache 2.2.

Apache 2.2 directives may be kept in 2.4, but may have "weird" unexpected behavior, so ifdef is the best solution.

eldk:
```
# Attempt to load mod_version if it wasn't loaded or compiled in (eg on Debian)
<IfModule !mod_version.c>
LoadModule version_module /usr/lib/httpd/modules/mod_version.so
</IfModule>
<Location /pagespeed_admin>
        <IfVersion <2.4>
            Order allow,deny
            Allow from localhost
            Allow from 127.0.0.1
            <IfVersion>
        <IfVersion >= 2.4>
            Require all granted
            Require local
            #Require ip, host or forward-dns  - see https://httpd.apache.org/docs/current/fr/mod/mod_authz_host.html
            </IfVersion>
        SetHandler pagespeed_admin
    </Location>
 <Location /pagespeed_global_admin>
        <IfVersion <2.4>
            Order allow,deny
            Allow from localhost
            Allow from 127.0.0.1
            <IfVersion>
        <IfVersion >= 2.4>
            Require all granted
            Require local
	     #Require ip, host or forward-dns
            </IfVersion>
          SetHandler pagespeed_global_admin
    </Location>
```

Is it ok that way ?
Thanks,
Eric

oschaaf:
@eldk do you stil plan to get this merged? If so, there are still a couple of outstanding review comments that need to be addressed before we can do so.

install/common/pagespeed.conf.template
```diff
@@ -320,17 +320,34 @@
...
+            Order allow,deny
+            Allow from localhost
+            Allow from 127.0.0.1
+            <IfVersion>
+        <IfVersion >= 2.4>
+            Require local
+            #see https://httpd.apache.org/docs/current/fr/mod/mod_authz_host.html - Require all granted,
+            #Require ip, Require host or Require forward-dns,
+            #if access must be granted to other than local.
+            </IfVersion>
```
oschaaf: nit: indentation

install/common/pagespeed.conf.template
```diff
@@ -320,17 +320,34 @@
...
-        Allow from 127.0.0.1
-        SetHandler pagespeed_global_admin
+        <IfVersion <2.4>
+            Order allow,deny
+            Allow from localhost
+            Allow from 127.0.0.1
+            <IfVersion>
+        <IfVersion >= 2.4>
+            Require local            
+            </IfVersion>
```
oschaaf: ditto

oschaaf (changes requested):
Two tiny comments about whitespace.
Other then that this looks good to me.
Did you test this?

eldk:
Hello @oschaaf 
>Did you test this?

I wasn't use it with version check as I don't use any Apache < 2.4.
I just verified it now with version check and corrected two errors : 
- missing space in `<IfVersion < 2.4>`
- two unclosed </IfVersion> container

Check is ok now : 
```
apache2ctl configtest 
Syntax OK
```

eldk:
I think it's good now ?

oschaaf (changes requested):

eldk:
I've got them.
Thanks for your patience,
Eric

oschaaf (approved):

PR #1707, opened 2018-01-10
Update build_ngx_pagespeed_from_source.html

oschaaf:
Update to make it work after the repo transfer


oppianmatt:
need to change later on also the nginx add-module line:

locally I went with:

```
cd incubator-pagespeed-ngx-${NPS_VERSION}/
```

instead of the wildcard

because later you have to specify that directory also in the add-module, don't think that takes a wildcard

```
--add-module=$HOME/incubator-pagespeed-ngx-${NPS_VERSION} 
```

but yes confirms the above now compiles and configures

oschaaf:
@oppianmatt updated to fix the ./configure line, could you take a another look?

oppianmatt:
builds and configures

PR #1711, opened 2018-01-11
New index for modpagespeed.com 

oschaaf:
Based on Jeff's earlier work on this, with a couple of changes:
- Link the examples that currently serve as the index as 
  http://staging.modpagespeed.com/examples/ in the new index
- Link the API documentation from the new index

Note: The html changes in this branch are staged at http://staging.modpagespeed.com/ to 
make life easier with regard to review.


oschaaf:
@jmarantz @morlovich WDYT?

oschaaf:
@jeffkaufman ping
(just a FYI, this PR is based on your earlier work on a new index for modpagespeed.com)

jeffkaufman:
Thanks for picking this up! I'm sorry I didn't finish it!

jmarantz:
+1 I think it's really great that you are picking this up, and organizationally it's a huge win.

I would like to see someone (not me) think about design for this though (not that current modpagespeed.com is any better).  But https://www.iiswebspeed.com/ looks a lot nicer.  Can we pick up some design structure from that?

oschaaf:
@jmarantz I'll ask around tomorrow to figure out what effort it would take to reuse some design elements. 
I do suspect it would involve touching all html pages, but it probably is a relatively small amount of actual work. So I will get back on that tomorrow.

As for this PR: would you be in favor of merging this as-is and swap staging. with www. today?
(I feel that doing so would be an improvement)

jmarantz (approved):
sure, lgtm

oschaaf:
done, new index is live!

PR #1715, opened 2018-01-13
Use madler/zlib.git, deprecate pagespeed/zlib

oschaaf:
Fix a TODO and deprecate our own version of zlib.
Use https://github.com/madler/zlib.git instead.


jmarantz (approved):
Seems fine to me, although I'm curious what makes the madler one authoritative?

oschaaf:
@jmarantz  See https://zlib.net/ and https://en.wikipedia.org/wiki/Mark_Adler

PR #1717, opened 2018-01-15
Update license banners in net/ and pagespeed/

oschaaf:
This updates license headers in net/ and pagespeed/ for .cc and .h files. The process
used to do this was: 

1. Strip the original banners with a tiny script
2. Run Apache RAT [1] to (re)insert banners into sources that do not have them
3. Perform some fixups, mostly needed to correct errors caused by the script in 1.

Note authors are also stripped from the comments in some cases. This is on purpose,
I believe ASF policy is that these should not be there, and there are arguments for not having
them. Unfortunately I have not been able to find an authoritative source on the subject yet.

While doing so was not the purpose of this change, a future change may strip the 
remaining author comments out.

[1] e.g. `java -jar ~/rat/apache-rat-0.13-SNAPSHOT.jar -e .rat-excludes -A --force -d net/`


oschaaf:
@zwoop do you know if there's an authorative source somewhere stating that authors shouldn't be included as comments in source code? I remember we don't do that, but I can't find it anywhere..

oschaaf:
@jukka see above, would you know?

net/instaweb/rewriter/public/data_url_input_resource.h
```diff
@@ -1,5 +1,3 @@
-#ifndef NET_INSTAWEB_REWRITER_PUBLIC_DATA_URL_INPUT_RESOURCE_H_
-#define NET_INSTAWEB_REWRITER_PUBLIC_DATA_URL_INPUT_RESOURCE_H_
 
 /*
  * Copyright 2010 Google Inc.
```
jukka: Should we update this header also?

jukka:
I don't think there's an official policy against author comments, but it's certainly a discouraged practice. The version control system is much better than author comments in tracking down people who know a particular piece of code best, and explicit author comments go against the principle of the project community being jointly responsible for the entire codebase. Some projects maintain a CONTRIBUTORS file or include thanks to all contributors in release notes, both of which practices are much better than author comments in giving kudos to contributors. None of these is necessary.

zwoop:
> On Jan 18, 2018, at 4:29 AM, Jukka Zitting <notifications@github.com> wrote:
> 
> I don't think there's an official policy against author comments, but it's certainly a discouraged practice. The version control system is much better than author comments in tracking down people who know a particular piece of code best, and explicit author comments go against the principle of the project community being jointly responsible for the entire codebase. Some projects maintain a CONTRIBUTORS file or include thanks to all contributors in release notes, both of which practices are much better than author comments in giving kudos to contributors. None of these is necessary.
> 
> 

Right. If I understand things, the NOTICE file is there to avoid having Copyright notices in sources files. So, for the cases where there was a Copyright notice in the source, and you removed it, likely it should be in the NOTICE file instead. However, we should not litter the NOTICE file with names / individuals if they are not necessary, it should only be used when it’s legally required.

See e.g.
	http://www.apache.org/legal/src-headers.html#notice <http://www.apache.org/legal/src-headers.html#notice>
	http://www.apache.org/dev/licensing-howto.html#mod-notice <http://www.apache.org/dev/licensing-howto.html#mod-notice>


So I guess it really depends. Most projects do not keep personal / individual attributions in the source files, and I assume as long as the previous license was clear, you can simply nuke them if you prefer and not add the individuals to the NOTICE file. However, anything that attributes a Copyright in the old source (like the one example with Google) you should add to the NOTICE file accordingly.

Cheers,

— Leif

oschaaf:
I started a [thread on the dev list](https://lists.apache.org/thread.html/4c984c8e3f1e75c08dc061385883dc82f648701d701ff5cd756903f2@%3Cdev.pagespeed.apache.org%3E
) to see wether there are any objections to stripping the author comments.

oschaaf:
Also -- thanks for the feedback @jukka & @zwoop. I'll amend accordingly and follow up when I did that.

oschaaf:
@zwoop / @jukka comments addressed, NOTICE initialized:
https://github.com/apache/incubator-pagespeed-mod/blob/oschaaf-licence-headers/NOTICE

If everyone is happy with the way this looks, I propose to merge this PR and strip the author: comments in a follow-up

PR #1718, opened 2018-01-16
Add Apache license banners in devel/ and html/

oschaaf:


jmarantz (commented):
Is it really necessary to add copyright headers to all the test html/js/css files?  Or we doing that for test images?

I thought maybe this might actually break some tests.

oschaaf:
@jmarantz you're right.. I've been looking in to figuring out how much it would take to fix the failures that resulted from this, and after some fiddling with it it looks like it may take quite some time.

There's a relevant section over at http://www.apache.org/legal/src-headers.html#faq-update-scripts [1]
I think we can (and I agree we should) exclude mod_pagespeed_test. I'll revert that part.

[1]

```
WHAT FILES IN AN APACHE RELEASE DO NOT REQUIRE A LICENSE HEADER?

A file without any degree of creativity in either its literal elements or its structure is not
protected by copyright law; therefore, such a file does not require a license header. If in doubt
about the extent of the  file's creativity, add the license header to the file.

Other files may make sense to have no license header. Three examples are:

- Short informational text files; for example README, INSTALL files. 
- The expectation is that these files make it obvious which product they relate to.
- Test data for which the addition of a source header would cause the tests to fail.

'Snippet' files that are combined as form a larger file where the larger file would have duplicate licensing headers.
PMCs should use their judgement, err on having a source header and contact legal-discuss@ if unsure.
```

jmarantz (approved):

oschaaf:
@jmarantz updated, this PR is now restricted to `devel/` and `html/`

PR #1724, opened 2018-01-18
Upgrading libpng library to version 1.6

ashishk-1:
Upgrading libpng library to version 1.6 for
- Bug fixes and performance enhancement as compared to version 1.2
- In case of Debian 9.3, this will help to dynamically link native libpng lib to pagespeed code

### DIFF
--- a/net/instaweb/rewriter/image.cc
+++ b/net/instaweb/rewriter/image.cc
@@ -21,6 +21,14 @@
 #include <algorithm>
 #include <cstddef>
 
+extern "C" {
+#ifdef USE_SYSTEM_ZLIB
+#include "zlib.h"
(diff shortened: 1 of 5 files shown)


morlovich (approved):
LGTM since:
1) mod_pagespeed.so uses a symbol map to only export pagespeed_module struct
2) ngx_pagespeed mode does scary symbol renaming
... So we don't risk linking with two different libpngs at same symbol names in normal build; and for dynamic builds this looks like it should be compatible back with 1.2 anyway.

oschaaf (approved):

oschaaf:
@ashishk-1 Checkin tests verified to pass, merging!

PR #1725, opened 2018-01-19
Fix a TODO, remove log_record.h forwarding header.

oschaaf:


jmarantz (approved):
would it be a big deal to re-order the include lines to keep them alphabetical?

oschaaf:
@jmarantz Ah.. no big deal at all, thanks for catching that. 
I sorted headers in a few more files in a follow-up [commit](https://github.com/apache/incubator-pagespeed-mod/pull/1725/commits/653d0f80241f7d12676fae3e543d047e18e05e22). Assuming no trouble arises with building the module, any objections?

PR #1739, opened 2018-02-03
Update .gitignore & .gitmodules 

oschaaf:
- Also, update .gitignore for log/ and release/
- Also, ignore some untracked content in external submodules to
  get a clean slate when running 'git status'

(this should fix running `install/build_release.sh --skip_psol --stable`)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ build/landmine_utils.pyc
 out/
 tools/closure/
 !devel/Makefile
+release/
+log/
--- a/.gitmodules
+++ b/.gitmodules
@@ -18,6 +18,7 @@
 	path = third_party/re2/src
 	url = https://github.com/google/re2.git
 [submodule "third_party/icu"]
+    ignore = untracked
 	path = third_party/icu
 	url = https://github.com/apache/incubator-pagespeed-icu.git
 [submodule "third_party/libjpeg_turbo/src"]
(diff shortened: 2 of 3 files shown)


oschaaf:
So it seems that the copy action is not always executed before `pnglibconf.h` is needed. The change I made in libpng.gyp does not fix that, because another build I had running just failed on it.

ashishk-1:
I will check this issue , when pnglibconf.h file is needed and when it is getting copied.

oschaaf:
@ashishk-1 great! So running 'make libpng` correctly executes the copy action correctly, but apparently that does not always run timely. Perhaps we need to add a dependency on libpng somewhere in gyp?
The build error I am currently looking at ends with:
```
  CXX(target) out/Release/obj.target/instaweb_spriter/net/instaweb/spriter/libpng_image_library.o
In file included from ./net/instaweb/spriter/libpng_image_library.h:32:0,
                 from net/instaweb/spriter/libpng_image_library.cc:20:
./third_party/libpng/src/png.h:361:27: fatal error: pnglibconf.h: No such file or directory
 #   include "pnglibconf.h"
                           ^
compilation terminated.
```

oschaaf:
Backed out my change to libpng.gyp & renamed this pull to indicate this only changes .gitignore and .gitmodules.

PR #1741, opened 2018-02-04
Update NOTICE and LICENSE

oschaaf:
On the devl list @morlovich mentioned: "The COPYRIGHT file in the Debian package may be of some use here, though it wouldn't include the 3rd part stuff that picks up from packages."

This is a first pass at assembling LICENSE and NOTICE, based on  https://github.com/apache/incubator-pagespeed-debian/blob/master/debian/copyright

If we can merge this, I think the next step would be to
- update https://docs.google.com/spreadsheets/d/1NTD2BKOEaX_8rQNKj8_6hdIbFl8UHYETfnZS2pkbvpw/edit#gid=0 
- cross off what is covered, see what is left to do and address that

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -187,16 +187,178 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2010 Google Inc.
 
-   Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


oschaaf:
Anyone up for reviewing / double checking this?

PR #1766, opened 2018-04-17
AddResourceHeader: always set. Ignore pre-existing values

oschaaf:


oschaaf:
@jmarantz @morlovich any objections to me merging this?

jmarantz:
Well, the directive does specifically indicate "Add", not "Set" or "Replace".  Can you just add a new directive "Set" with this behavior?

I can imagine there are some scenarios where "Add" makes sense, such as Cache-Control or Accept.

net/instaweb/rewriter/resource_fetch.cc
```diff
@@ -175,6 +175,7 @@ void ResourceFetch::HandleHeadersComplete() {
   for (int i = 0; i < driver_->options()->num_resource_headers(); ++i) {
     const RewriteOptions::NameValue* nv =
         driver_->options()->resource_header(i);
+    response_headers()->RemoveAll(nv->name);
     response_headers()->Add(nv->name, nv->value);
```
morlovich, line 179: Just using ->Replace might be a little nicer. Not that it's implemented any different

morlovich (commented):
So this is kind of missing explanation for why you would do this in the CL description; if only by linking to the bug report or support thread that motivated it --- though also summarizing would probably be better since I seem to remember a long conversation.

oschaaf:
@jmarantz absolutely. I'll amend. @morlovich looks nicer indeed, consider it done.

PR #1774, opened 2018-06-01
Add <code> to sensitive tags in collapse_whitespace_filter.

### LINKED ISSUE #1772: collapse_whitespace in <code>
The documentation says:

> ..is normally ignored by the browser outside of tags like &lt;pre> and <textarea>, one can use CSS properties such as "white-space: pre" to make the browser preserve whitespace within a portion of the document..

It would be nice if collapse_whitespace would also ignore &lt;code> tags, as these are quite a common case for "white-space: pre".

Additionally, MDN says:

> The HTML &lt;code> element displays its contents styled in a fashion intended to indicate that the text is a short fragment of computer code.

Based on that description, it's plausible that a browser may choose to pay attention to whitespace, where it normally wouldn't (given that, for example, Python code without the whitespace would be invalid code).

So, skipping &lt;code> content would make this filter a fair bit safer to use.

Dreamsorcerer:
Fixes #1772

### DIFF
--- a/pagespeed/kernel/html/collapse_whitespace_filter.cc
+++ b/pagespeed/kernel/html/collapse_whitespace_filter.cc
@@ -39,7 +39,7 @@ namespace {
 // Tags within which we should never try to collapse whitespace (note that this
 // is not _quite_ the same thing as kLiteralTags in html_lexer.cc):
 const HtmlName::Keyword kSensitiveTags[] = {
-  HtmlName::kPre, HtmlName::kScript, HtmlName::kStyle, HtmlName::kTextarea
+  HtmlName::kCode, HtmlName::kPre, HtmlName::kScript, HtmlName::kStyle, HtmlName::kTextarea
 };
 
 bool IsSensitiveKeyword(HtmlName::Keyword keyword) {
--- a/pagespeed/kernel/html/collapse_whitespace_filter_test.cc
+++ b/pagespeed/kernel/html/collapse_whitespace_filter_test.cc
@@ -62,6 +62,12 @@ TEST_F(CollapseWhitespaceFilterTest, NewlineTakesPrecedence) {
                    "is good to see you</body>");
 }
 
+TEST_F(CollapseWhitespaceFilterTest, DoNotCollapseWithinCode) {
+  ValidateNoChanges("do_not_collapse_within_code",
+                    "<body><code>hello   world,   it\n"
+                    "    is good  to     see you   </code></body>");
+}
+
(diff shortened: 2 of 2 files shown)


pagespeed/kernel/html/collapse_whitespace_filter.cc
```diff
@@ -39,7 +39,7 @@ namespace {
 // Tags within which we should never try to collapse whitespace (note that this
 // is not _quite_ the same thing as kLiteralTags in html_lexer.cc):
 const HtmlName::Keyword kSensitiveTags[] = {
-  HtmlName::kPre, HtmlName::kScript, HtmlName::kStyle, HtmlName::kTextarea
+  HtmlName::kCode, HtmlName::kPre, HtmlName::kScript, HtmlName::kStyle, HtmlName::kTextarea
```
jmarantz, line 42: very cool.  Could you also add a test in collapse_whitespace_filter_test.cc?  The pattern should be fairly obvious there.

oschaaf:
Nice, thanks!

jmarantz:
@oschaaf do you know why CI is failing?  I re-started it once, but both attempts ended in:

ld/apache/incubator-pagespeed-mod/third_party/giflib/src'...
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated

I'll try a third time. If that fails we can I guess patch/build/test and then merge, but I wanted to see if you thought the CI was failing like this generally and if we should so something about that.

oschaaf:
@jmarantz I noticed the same with another PR lately, but have not investigated yet. perhaps one of the dependencies broke, I’ll take a look later today

oschaaf:
So the test failed on `But do proxy content if the Content-Type header is present.`.
This was very reproducible because a required script that serves test responses was not running on modpagespeed.com. I recently rebuild the machine from scrap to simplify some things, and forgot all about that script. Restarted the CI job, expect this to be fixed.
(Though sometimes CI may still flake because of the tests requiring an amount of time that is pretty close to the timeout limit.)

oschaaf:
@jmarantz +1 for merging

With regard to giflib, that started failing when sourceforge starting migrating
datacenters. I already attempted a fix [1] earlier because the previous location started 
responding with 404 .. 
But apparently the git repo we now point to times out every now and then, ugh.

[1] https://github.com/apache/incubator-pagespeed-mod/commit/8196fc74847e85085fa6f310d658842261acb329

PR #1776, opened 2018-06-07
Redis: Use Redis DB-Index when computing the SystemCache lookup key

### LINKED ISSUE #1771: Redis DB index aliasing
Looking at the code, I believe that if two different VirtualHosts in the same config construct redis caches that differ only DB-index then the first one will win, and the second db-index will be ignored.

n #1627 I think the db-index needs to be concatenated into the key used to lookup cache objects to be shared between multiple servers, in this block code in system_caches.cc:

```
    spec_signature =
        StrCat("r;", config->redis_server().ToString(), ";",
               IntegerToString(config->redis_reconnection_delay_ms()), ";",
               IntegerToString(config->redis_timeout_us()));
```

oschaaf:
The added test didn't turn out the way I planned to because of running into some unrelated
things (see the comment in the test script). 
The full checkin tests are still running, but I did observe the added test itself to pass.

Should fix https://github.com/apache/incubator-pagespeed-mod/issues/1771

### DIFF
--- a/install/debug.conf.template
+++ b/install/debug.conf.template
@@ -28,6 +28,7 @@ AddOutputFilterByType MOD_PAGESPEED_OUTPUT_FILTER application/xhtml+xml
 #REDIS
 #REDIS # Do testing using redis instead of file cache
 #REDIS ModPagespeedRedisServer localhost:@@REDIS_PORT@@
+#REDIS ModPagespeedRedisDatabaseIndex 2
 #EXTCACHE
 # If X-PSA-Blocking-Rewrite request header is present and its value matches the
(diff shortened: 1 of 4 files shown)


oschaaf:
(off course, CI is till broken, working on figuring out how to fix that)

oschaaf:
(rebased this PR so we can hopefully observe how CI feels about this change )

oschaaf:
checkin tests not passing. It seems that REDIS_PORT is set as an environment variable, when we're not actually running the tests with redis. (I was assuming that wouldn't be the case)

oschaaf:
Merging this as-is in to solve this issue, I will file a separate issue on the REDIS_PORT being set when tests are run without redis.

PR #1777, opened 2018-06-07
CI: fix hangs

oschaaf:
The old https:// source of giflib is finally back up again after they moved datacenters. So let's point the submodule back to that. Also, let's reduce the amount of parallel submodule fetching as to not anger any rate limiting systems. Let's see if we'll be able to run the tests within timeout limit that we're given with these changes.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -114,4 +114,5 @@
 	url = https://github.com/madler/zlib.git
 [submodule "third_party/giflib/src"]
 	path = third_party/giflib/src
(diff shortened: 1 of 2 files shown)


jmarantz (approved):

oschaaf:
So, the good news is that the git timeouts are resolved. The bad news is that now Travis times out.
So I'm not done with getting CI back up yet

oschaaf:
Experimented with running `git-cache-http-server`, hosting it on GCE. 
Unfortunately the experiment didn't work out very well.

PR #1778, opened 2018-06-08
CentOS: changed urls for sl devtoolset repo

oschaaf:
This needs to be confirmed working, but we probably need this.

(reported via  https://groups.google.com/d/msgid/mod-pagespeed-discuss/f1ed9698-da93-4792-a857-a1a5707e539a%40googlegroups.com)


oschaaf:
Manually confirmed that `install_required_packages.sh` is working OK with this change on a clean CentOS 6 VM, merging

PR #1785, opened 2018-07-01
Fix mismatch between decompression and headers

morlovich:
Fix mismatch between headers and demand-uncompressed payload when hitting rewrite deadline with gzip'ed input.

Believed to fix issue 1556

### DIFF
--- a/net/instaweb/rewriter/resource.cc
+++ b/net/instaweb/rewriter/resource.cc
@@ -62,7 +62,7 @@ Resource::Resource(const RewriteDriver* driver, const ContentType* type)
       disable_rewrite_on_no_transform_(true),
       is_authorized_domain_(true),
       respect_vary_(ResponseHeaders::kRespectVaryOnResources),
-      extracted_(false) {
+      extracted_state_(kExtractNotComputed) {
 }
 
 Resource::Resource() : server_context_(NULL), type_(NULL),
@@ -74,7 +74,7 @@ Resource::Resource() : server_context_(NULL), type_(NULL),
                        disable_rewrite_on_no_transform_(true),
                        is_authorized_domain_(true),
                        respect_vary_(ResponseHeaders::kRespectVaryOnResources),
-                       extracted_(false) {
+                       extracted_state_(kExtractNotComputed)  {
 }
 
 Resource::~Resource() {
@@ -256,8 +256,9 @@ bool Resource::Link(HTTPValue* value, MessageHandler* handler) {
   DCHECK(UseHttpCache());
   const SharedString& contents_and_headers = value->share();
   // Invalidate extracted_contents_.
-  extracted_ = false;
+  extracted_state_ = kExtractNotComputed;
   extracted_contents_.clear();
+  extracted_headers_ = nullptr;
   return value_.Link(contents_and_headers, &response_headers_, handler);
 }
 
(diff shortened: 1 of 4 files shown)


morlovich:
(Note that I've only ran this through unit tests...)
The high risk point is probably using the headers in response_headers_ rather than value_ --- when are they out-of-sync?

There are also a few other spots that potentially need the same treatment, not sure if to bundle them in.

oschaaf:
Awesome! This passes `apache_debug_smoke_test`, but the `checkin` tests fail when running `apache_debug_slurp_test`. 

I looked into that, and this happens because of this needing to be rebased onto master, which has changes to reflect changes to modpagespeed.com's structure.
I will take a closer look at the changes later today.

net/instaweb/rewriter/resource.cc
```diff
@@ -269,17 +270,38 @@ void Resource::LinkFallbackValue(HTTPValue* value) {
...
+  return use_extracted ? extracted_contents_ : raw_contents();
+}
+
+const ResponseHeaders* Resource::UncompressedHeaders() const {
+  bool use_extracted = EnsureExtractedIfNeeded();
+  return use_extracted ? extracted_headers_.get() : response_headers();
+}
+
+bool Resource::EnsureExtractedIfNeeded() const {
+  if (extracted_state_ == kExtractNotComputed) {
```
oschaaf, line 283: I wonder what happens when `response_headers_` gets changed after `extracted_state_` is computed.
If that is a valid case, perhaps this needs to reset `extracted_state_` to `kExtractNotComputed` when we observe response_headers_ is dirty / has been modified?

oschaaf (approved):
One question, but that aside this looks great to me

jmarantz:
Still haven't been able to dive into the code, but I wonder if anyone who
is experiencing the broken behavior would be willing to patch in this PR
and give it a spin.

net/instaweb/rewriter/resource.cc
oschaaf: Giflib is hosted on sourceforge, which seems to be very unstable in the last couple of months. Maybe we should set up a mirror for giflib on github.

morlovich: It's actually not that but rather the submodule registration stuff --- missing pngconf.h or such somehow.

oschaaf: Ugh. Manually cp'ing third_party/libpng/src/scripts/pnglibconf.h.prebuilt to third_party/libpng/src/pnglibconf.h should work.

This happens because a build execution step isn't always executed timely. I run make with `-j8`  which seems to mask the problem. But with -j1 the problem does show on my system.


In https://github.com/apache/incubator-pagespeed-mod/pull/1739 I backed out two commits that attempted to fix that to avoid conflicting with Ashish his work:
- https://github.com/apache/incubator-pagespeed-mod/pull/1739/commits/2b4ae6032986654dc73ceee48e9f70de9267a9ad
- https://github.com/apache/incubator-pagespeed-mod/pull/1739/commits/f72e2111070dc07030e0390dfe07f7ae65d78826

Let me try with those again to see if they fix the problem.

oschaaf: So giflib was fixed, CI is passing again on master, let me see if I can fix the copying of third_party/libpng/src/scripts/pnglibconf.h as well

oschaaf: Ugh, I'm having a real problem fixing the problem with libpng failing on single-threaded builds. I have been at it all day, but no luck yet.

Apparently there is an ordering issue that gets masked when using -j in make. The problem is that the dependency declarations for libpng look good to me, yet somehow the action [1] responsible for copying a required header file isn't always running timely when it ought to.

[1] https://github.com/apache/incubator-pagespeed-mod/blob/master/third_party/libpng/libpng.gyp#L27

oschaaf:
@morlovich tests are passing, are you ok with merging this?

oschaaf:
As discussed outside of this thread, merged this.

PR #1791, opened 2018-08-07
Use giflib mirror on github

oschaaf:


jmarantz (approved):
I'm fine either way, but why not put this under githob.com/apache/incuabor-pagespeed-mod/ ?

License issue?

oschaaf:
I think there is no license issue, but mirroring the entire giflib repo in ours wouldn’t be possible with history and all? So we I thought we needed a new repo to mirror this, and requesting incubating-pagespeed-giflib-mirror.git or something like that didn’t make sense to me.. (though it used to be mirrored under the pagespeed/ git org which had a slight modification for gyp which is no longer needed).

jmarantz:
sounds good.

PR #1802, opened 2018-10-14
Prefer Paint Timing API over chrome.loadTimes()

### LINKED ISSUE #1537: DNS prefetch function uses "chrome.loadTimes" instead of the API Paint Timing
Hi,
i updated to `Chrome 64` and from the console in my website i receive an advice as the follow:
`[Deprecation] chrome.loadTimes() is deprecated, instead use standardized API: Paint Timing.`

I found the problem in mod_pagespeed, surely it's in it. I think it's the "insert_dns_prefetch" filter but i'm note sure of this filter. Seems that it uses `chrome.loadTimes` instead of the new API `Paint Timing`. That should be the problem.

Could you confirm that? I've `mod_pagespeed 1.11.33.4` installed.

Thanks

injust:
Fixes apache/incubator-pagespeed-ngx#1537

### DIFF
--- a/net/instaweb/rewriter/add_instrumentation.js
+++ b/net/instaweb/rewriter/add_instrumentation.js
@@ -71,7 +71,8 @@ pagespeed.AddInstrumentation.prototype.sendBeacon = function() {
   }
 
   url += '&r' + this.event_ + '=';
-  var winPerf, timingApi;
+  var winPerf, timingApi, paintEntries, loadTimes;
+  var firstPaintTime = -1;
   if ((winPerf = window['performance']) && (timingApi = winPerf['timing'])) {
     var navStartTime = timingApi['navigationStart'];
     var requestStartTime = timingApi['requestStart'];
@@ -91,8 +92,6 @@ pagespeed.AddInstrumentation.prototype.sendBeacon = function() {
     if (winPerf['navigation']) {
       url += '&nt=' + winPerf['navigation']['type'];
     }
-    var firstPaintTime = -1;
-    var paintEntries, loadTimes;
     if (timingApi['msFirstPaint']) {
       // IE.
       firstPaintTime = timingApi['msFirstPaint'];
--- a/net/instaweb/rewriter/add_instrumentation.js
+++ b/net/instaweb/rewriter/add_instrumentation.js
@@ -96,12 +96,13 @@ pagespeed.AddInstrumentation.prototype.sendBeacon = function() {
       // IE.
       firstPaintTime = timingApi['msFirstPaint'];
     } else if (window['performance']['timeOrigin'] &&
-        window['performance']['getEntriesByName'] &&
-        window['performance']['getEntriesByName']('first-paint').length > 0) {
+        window['performance']['getEntriesByType'] &&
+        window['performance']['getEntriesByType']('paint').length > 0) {
         // Paint Timing API
+        // Note that getEntriesByType('paint') is sorted chronologically
         firstPaintTime = Math.floor(
             window['performance']['timeOrigin'] +
-            window['performance']['getEntriesByName']('first-paint')[0]['startTime']);
+            window['performance']['getEntriesByType']('paint')[0]['startTime']);
     } else if (window['chrome'] && window['chrome']['loadTimes']) {
       // Chrome. Note that window.chrome.loadTimes returns a time in seconds.
       firstPaintTime = Math.floor(
--- a/net/instaweb/rewriter/add_instrumentation.js
+++ b/net/instaweb/rewriter/add_instrumentation.js
@@ -100,12 +100,12 @@ pagespeed.AddInstrumentation.prototype.sendBeacon = function() {
         (paintEntries = winPerf['getEntriesByType']('paint')).length > 0) {
         // Paint Timing API
         // Note that getEntriesByType('paint') is sorted chronologically
-        firstPaintTime = Math.floor(
+        firstPaintTime = Math.round(
             winPerf['timeOrigin'] +
             paintEntries[0]['startTime']);
     } else if ((loadTimes = window['chrome']) && (loadTimes = loadTimes['loadTimes'])) {
       // Chrome. Note that window.chrome.loadTimes returns a time in seconds.
-      firstPaintTime = Math.floor(
+      firstPaintTime = Math.round(
           loadTimes()['firstPaintTime'] * 1000);
     }
     timeToFirstPaint = firstPaintTime - requestStartTime;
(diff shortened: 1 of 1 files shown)


net/instaweb/rewriter/add_instrumentation.js
```diff
@@ -95,6 +95,12 @@ pagespeed.AddInstrumentation.prototype.sendBeacon = function() {
     if (timingApi['msFirstPaint']) {
       // IE.
       firstPaintTime = timingApi['msFirstPaint'];
+    } else if (window['performance']['timeOrigin'] &&
```
jmarantz: Just for a bit of paranoia, could you make sure getEntriesByName is non-null before calling it?

net/instaweb/rewriter/add_instrumentation.js
```diff
@@ -95,6 +95,12 @@ pagespeed.AddInstrumentation.prototype.sendBeacon = function() {
...
+        window['performance']['getEntriesByName']('first-paint').length > 0) {
+        // Paint Timing API
+        firstPaintTime = Math.floor(
+            window.performance.timeOrigin +
```
jmarantz: Use the ['xxx'] notation rather than .xxx  as above. This is important because the Closure Compiler may rename these.

net/instaweb/rewriter/add_instrumentation.js
```diff
@@ -95,6 +95,12 @@ pagespeed.AddInstrumentation.prototype.sendBeacon = function() {
...
+            window.performance.getEntriesByName('first-paint')[0].startTime);
```
jmarantz: I'm a bit OCD when it comes to CSE :)  WDYT of making a temp `winPerf` for window['performance'] and then using it afterward, and likeways for winPerf['getEntriesByName'] and others where you test first and then call, or have chains of dereferencing...

This is just a suggestion, and it would require something like a nested helper function to help you manage the temps in the conditional.

jmarantz (commented):
Were you able to rebuild mod_pagespeed with this change and try it out? DId it fix the problem for you with the deprecation warnings?

net/instaweb/rewriter/add_instrumentation.js
injust: I'm not sure how I would use a helper function for preventing chains of dereferencing, could you clarify that bit?
I gave it a shot with temporary variables, but feel like there's a better way of doing it.

net/instaweb/rewriter/add_instrumentation.js
```diff
@@ -71,11 +71,11 @@ pagespeed.AddInstrumentation.prototype.sendBeacon = function() {
   }
 
   url += '&r' + this.event_ + '=';
-  if (window['performance']) {
-    var timingApi = window['performance']['timing'];
+  if (winPerf = window['performance']) {
```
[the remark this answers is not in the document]
injust: Right, oops.
It compiles, but bad practice to use undeclared variables. Sorry!

oschaaf:
This looks good to me; did you get a chance to test this?
Is this ready for merging?

net/instaweb/rewriter/add_instrumentation.js
```diff
@@ -88,21 +88,30 @@ pagespeed.AddInstrumentation.prototype.sendBeacon = function() {
...
+    if (winPerf['navigation']) {
+      url += '&nt=' + winPerf['navigation']['type'];
     }
     var firstPaintTime = -1;
+    var paintEntries, loadTimes;
     if (timingApi['msFirstPaint']) {
       // IE.
       firstPaintTime = timingApi['msFirstPaint'];
-    } else if (window['chrome'] && window['chrome']['loadTimes']) {
+    } else if (winPerf['timeOrigin'] &&
```
jmarantz: can you do the same trick here -- predeclaring vars for timeOrigin and getEntriesByType and then assigning them in the if-statement as side effects. That looks pretty readable.

net/instaweb/rewriter/add_instrumentation.js
```diff
@@ -88,21 +88,30 @@ pagespeed.AddInstrumentation.prototype.sendBeacon = function() {
...
       // Chrome. Note that window.chrome.loadTimes returns a time in seconds.
-      firstPaintTime = Math.floor(
-          window['chrome']['loadTimes']()['firstPaintTime'] * 1000);
+      firstPaintTime = Math.round(
+          loadTimes()['firstPaintTime'] * 1000);
     }
-    firstPaintTime = firstPaintTime - requestStartTime;
-    if (firstPaintTime >= 0) {
-      url += '&fp=' + firstPaintTime;
+    timeToFirstPaint = firstPaintTime - requestStartTime;
```
jmarantz: var

net/instaweb/rewriter/add_instrumentation.js
```diff
@@ -88,21 +88,30 @@ pagespeed.AddInstrumentation.prototype.sendBeacon = function() {
...
     if (timingApi['msFirstPaint']) {
       // IE.
       firstPaintTime = timingApi['msFirstPaint'];
-    } else if (window['chrome'] && window['chrome']['loadTimes']) {
+    } else if (winPerf['timeOrigin'] &&
+        winPerf['getEntriesByType'] &&
+        (paintEntries = winPerf['getEntriesByType']('paint')).length > 0) {
+        // Paint Timing API
+        // Note that getEntriesByType('paint') is sorted chronologically
+        firstPaintTime = Math.round(
```
jmarantz: can you move the var declaration for firstPaintTime outside the if-block so it's obvious that scopes to all the branches?

jmarantz (commented):
Thanks for doing this -- this is great stuff so far.

To clarify my previous question -- when I asked whether this compiles -- I meant in the mod_pagepseed build process.

What happens in the build process is that this file gets built by a javascript compiler called 'closure compiler' into a faster/smaller version of itself. The resultant JS file is then compiled into a C++ source file that represents the code as C strings, so it can be served directly from the module without having to find it on disk.

This whole flow is automated by the mod_pagespeed build system. But it looks to me like some of the JS you have here might not make the Closure Compiler happy (missing var declarations mainly).

I also don't know how you could test this fully without doing that. But if you actually *have* done that -- and I'm wrong about the missing var declarations causing the compile to fail -- then apologies :)

injust:
Just tested with the mod_pagespeed build process. Indeed, Closure Compiler doesn't like the omission of `var` declarations.

Right now, `getEntriesByType('paint')` always returns an empty array. The entries only populate after the `load` event for `window`. Is it possible to defer running the entire instrumentation block?

I noticed that the build process uses pre-optimized files in [`net/instaweb/genfiles/rewriter`](https://github.com/apache/incubator-pagespeed-mod/tree/master/net/instaweb/genfiles/rewriter), which will need to be updated. How were those initially generated?

oschaaf:
@injust is this good to go as far as you are concerned?

PR #1854, opened 2019-01-28
Added support for specifying a TTL for redis keys (Fixes #1704)

### LINKED ISSUE #1704: Redis cache TTL
I've had a site with PageSpeed and Redis cache running for a few months now, and when running `redis-cli INFO` I've noticed this:

```
db1:keys=840116,expires=0,avg_ttl=0
```

It seems like PageSpeed doesn't set expire for the shared memory cache, keeping lots of unnecessary keys in the cache. Is there a way to set the expire time?

mtorromeo:
This is an attempt at implementing support for setting an expiration time to redis keys.
Issue #1704 describes the problem that this is trying to solve.

I have not developed in C++ in years and am not too familiar with the pagespeed codebase but I hope this is fine.

### DIFF
--- a/pagespeed/system/system_rewrite_options.cc
+++ b/pagespeed/system/system_rewrite_options.cc
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
@@ -37,6 +37,7 @@ namespace {
 
 const int64 kDefaultCacheFlushIntervalSec = 5;
 const int64 kDefaultRedisDatabaseIndex = 0;
+const int64 kDefaultRedisTTLSec = -1;
 
 const char kFetchHttps[] = "FetchHttps";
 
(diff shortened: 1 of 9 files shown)


jmarantz (commented):
Basically looks good except for some nits one significant omission. See https://github.com/apache/incubator-pagespeed-mod/blob/master/pagespeed/system/system_caches.cc#L288

The module keeps a map of RedisCache instances, which is keyed by a string composed of all the config options that are used to create the redis cache instance. You need to add your new TTL field into that string.

mtorromeo:
I should have applied all the changes.
I also fixed a typo in the rewrite option name and an incorrect default of `0` for the ttl.

oschaaf:
Awesome. It would be nice to have a system test for this, to make sure the feature doesn't get broken.
But I would be happy to merge this as-is. @mtorromeo wdyt?

pagespeed/system/system_rewrite_options.cc
```diff
@@ -55,6 +56,8 @@ const char SystemRewriteOptions::kRedisReconnectionDelayMs[] =
 const char SystemRewriteOptions::kRedisTimeoutUs[] = "RedisTimeoutUs";
 const char SystemRewriteOptions::kRedisDatabaseIndex[] =
     "RedisDatabaseIndex";
+const char SystemRewriteOptions::kRedisTTLSec[] =
+    "RedisTTLSec";
```
jmarantz: fits on one line?

jmarantz (commented):
You can, for no cost, get VirtualBox installed and an Ubuntu image and run tests on it.

Everything basically looks great but I agree with Otto it would be very nice to have tests for it.

oschaaf:
I propose we merge this, but leave https://github.com/apache/incubator-pagespeed-mod/issues/1704 open and make a remark there that the feature is there but needs test coverage.

jmarantz:
ok, sgtm.

PR #1886, opened 2019-05-13
remove id attributes from styles

Lofesa:
To solve this issue: prioritize critical css need to remove id attributes apache#1878
When inlined styles have ID´s if prioritize critical css filter is used, when pagespeed restores the inlined styles from class name psa_add_styles, the ID´s become duplicate.
The changes in this js file deletes the ID´s from these restored styles.
These change is not the best approach, i know, cause it add more work in the browser side, maybe some user use the ID´s to make changes to the DOM and maybe the elements the user will change don´t exists, cause the styles with the ID´s are the optimized ones.
But this is a workaround for these user that don´t do changes in styles DOM by the ID. 
Wordpress put an ID in all inlined styles by the hook wp_add_inline_style( $handle, $data ) where the $handle become the ID and is mandatory.

### DIFF
--- a/net/instaweb/rewriter/critical_css_loader.js
+++ b/net/instaweb/rewriter/critical_css_loader.js
@@ -42,12 +42,12 @@ pagespeed.CriticalCssLoader.addAllStyles = function() {
     
   for (var i = 0, e; e = elements[i]; ++i) {
     if (e.nodeName == 'NOSCRIPT') { 
-       var div = document.createElement('div');
-       div.innerHTML = e.textContent;
-       var children = div.childNodes;
-       for (var v = 0; v < children.length; ++v) {
-           children[v].removeAttribute('id');
-       }  
+      var div = document.createElement('div');
+      div.innerHTML = e.textContent;
+      var children = div.childNodes;
+      for (var v = 0; v < children.length; ++v) {
+    children[v].removeAttribute('id');
+    }  
     document.body.appendChild(div);
     }
  }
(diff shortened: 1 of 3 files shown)


net/instaweb/rewriter/critical_css_loader.js
```diff
@@ -39,12 +39,18 @@ pagespeed.CriticalCssLoader.addAllStyles = function() {
...
-    var div = document.createElement('div');
-    div.innerHTML = e.textContent;
+    if (e.nodeName == 'NOSCRIPT') { 
+       var div = document.createElement('div');
+       div.innerHTML = e.textContent;
+       var children = div.childNodes;
+       for (var v = 0; v < children.length; ++v) {
+           children[v].removeAttribute('id');
+       }  
     document.body.appendChild(div);
```
oschaaf: Nit indentation

oschaaf (changes requested):
Awesome; this LGTM module one style nit. The CI failure looks unrelated, one of the tests needs to update its expectations.

net/instaweb/rewriter/critical_css_loader.js
Lofesa: Indentation spaces changed by tabs

oschaaf: Thanks; I proposed one more whitespace change in a4b0e7b, wdyt?

Lofesa: You are rigth, I made mistakes when correcting, sorry for that.

oschaaf (approved):
Looks good to me, thanks!

oschaaf:
@jmarantz any objections to merging this?

PR #1888, opened 2019-05-16
Remove /html

oschaaf:
The source of truth for the project website now lives over at
https://github.com/apache/incubator-pagespeed-website


oschaaf:
DO NOT MERGE YET
We should only merge this after pagespeed.incubator.apache.org reflects the content from https://github.com/apache/incubator-pagespeed-website

oschaaf:
http://pagespeed.incubator.apache.org/ is now live, but I think it's a good idea and possible to keep the source of truth in this repository. I'll add a publish script we can run, and close this.

PR #1890, opened 2019-05-16
Update GRPC dependency: 1.6.1->1.6.8

oschaaf:


oschaaf:
This is the highest grpc version we can go to without the build breaking; it's probably worth upgrading
to the latest and greatest at some point though.

Issue #1849, opened 2018-12-19
Add srcset as a possible category for UrlValuedAttribute


weberdigital:
In some cases, we need to put srcsets for image into a custom attribute, such as data-srcset. 

Unfortunately, those images cannot be optimized with ModPagespeed at the moment. Adding the configuration _ModPagespeedUrlValuedAttribute img data-srcset image_ does not work, as data-srcset does include both the image url as well as the image size. Thus, it would be nice if there was a dedicated category for srcsets in for the UrlValuedAttributes.

Lofesa:
Hi @oschaaf 
Yes, is a improvement, like data-src work as img category. 
As you know, I´m not a developer but I think most of the work is done as is the same issue as srcset label.
Perhaps an other issue/improvement with the lazyload_images filter because it load images in onload event, no matter if LazyloadImagesAfterOnload is set to on or off and Pagespeed Insigths don´t recognize that images are lazyloaded.
Because of that I´m using a wordpress plugin based on [lazysizes ](https://github.com/aFarkas/lazysizes) that need data-src and data-srcset

andymathieson:
Hi @oschaaf. 
Brilliant that you seem to be working on this. Does it require further testing?

Lofesa:
Hi @andymathieson 
Sure, none of us take care of all use cases. 
This PR is a transient one. The definitive PR is pending to commit. I have it in production, so it works. But my site is a small web, maybe some use cases are not covered.

andymathieson:
@Lofesa 
Thanks for the reply. We require the data-srcset capabilty so will probably fire up a test server and give it a spin. Hopefully will be commited to master soon as I'm sure this is a highly needed feature.

Lofesa:
Hi @andymathieson 
If you fire a test server, apply the PR #1929 , the definitive pacht to solve the issue.
This PR solves data-srcset, amp img srcset and add a UrlValuedAttibute srcset category, so you can use others elements:
ModPagespeedUrlValuedAttribute img data-wpfc-original-srcset srcset

andymathieson:
Hi @Lofesa 
Will this patch be merged into the master anytime soon in the future?

Lofesa:
Hi @andymathieson 
I don´t know when the patch be merged and when a new release that include it be libered.
Maybe @oschaaf can say some about this.

Jacobs63:
It's been almost a year without any response.

Any updates? @oschaaf

[...]

Issue #1608, opened 2017-07-29
Dublicate Content Error because of Url : PageSpeed=noscript 


oguzcicek:
Hello , 
I have the problem of duplicate Meta tag ,  duplicate Meta title because of 
html?PageSpeed=noscript  

Should i add this to robot.txt ? 
Disallow: /*?PageSpeed=noscript 

How can i handle this problem ?

oschaaf:
Embedding the canonical url in the html may help:
https://support.google.com/webmasters/answer/139066?hl=en

Alternatively, you could disable the filters that depend on javascript to render the page correctly, or leave the filters enabled and drop support for clients that do not have javascript enabled. 
(See https://modpagespeed.com/doc/faq#noscript-redirect)

PR #1899, opened 2019-06-16
data-srcset in img and amp-img

Lofesa:
To rewrite amp-img srcset the parameter pagespeed UrlValuedAttribute amp-img  src image; is needed.
In fact, i think that any UrlValuedAttribute element attribute image that have src,srcset or data-srcset get rewrited.
Solves [#1608 ](https://github.com/apache/incubator-pagespeed-ngx/issues/1608)in nginx and #1849.

### DIFF
--- a/net/instaweb/rewriter/image_rewrite_filter.cc
+++ b/net/instaweb/rewriter/image_rewrite_filter.cc
@@ -2138,12 +2138,14 @@ void ImageRewriteFilter::StartElementImpl(HtmlElement* element) {
     BeginRewriteImageUrl(element, attributes[i].url);
    
   
-    HtmlElement::Attribute* srcset[] = {element->FindAttribute(HtmlName::kSrcset),element->FindAttribute(HtmlName::kDataSrcset)};
-    for (int i = 0; i == 1; i++){
-       if (srcset[i] != nullptr){
-	       BeginRewriteSrcSet(element, srcset[i]);
-        }
-    }
+    HtmlElement::Attribute* srcset = element->FindAttribute(HtmlName::kSrcset);
+    HtmlElement::Attribute* datasrcset = element->FindAttribute(HtmlName::kDataSrcset);
+    if (srcset != nullptr) {
+	    BeginRewriteSrcSet(element, srcset);
+    } else if (datasrcset != nullptr) {
+	    BeginRewriteSrcSet(element, datasrcset);
+    }	    
+  
   }
 }
 
--- a/net/instaweb/rewriter/image_rewrite_filter.cc
+++ b/net/instaweb/rewriter/image_rewrite_filter.cc
@@ -2139,7 +2139,7 @@ void ImageRewriteFilter::StartElementImpl(HtmlElement* element) {
    
   
     HtmlElement::Attribute* srcset[0] = element->FindAttribute(HtmlName::kSrcset);
-    HtmlElement::Attribute* srcset[1] = element->FindAttribute(HtmlName::kDataSrcset);
+    srcset[1] = element->FindAttribute(HtmlName::kDataSrcset);
     for (int i = 0; i == 1; i++){
        if (srcset[i] != nullptr){
 	       BeginRewriteSrcSet(element, srcset[i]);
--- a/net/instaweb/rewriter/image_rewrite_filter.cc
+++ b/net/instaweb/rewriter/image_rewrite_filter.cc
@@ -2136,13 +2136,18 @@ void ImageRewriteFilter::StartElementImpl(HtmlElement* element) {
     }
 
     BeginRewriteImageUrl(element, attributes[i].url);
-  }
-
-  if (element->keyword() == HtmlName::kImg) {
+   
+  
     HtmlElement::Attribute* srcset = element->FindAttribute(HtmlName::kSrcset);
+	  HtmlElement::Attribute* datasrcset = element->FindAttribute(HtmlName::kDataSrcset);
     if (srcset != nullptr) {
       BeginRewriteSrcSet(element, srcset);
     }
+	  if (datasrcset != nullptr) {
+      srcset = datasrcset ;
+      BeginRewriteSrcSet(element, srcset);
+    }
+
   }
 }
 
--- a/net/instaweb/rewriter/image_rewrite_filter.cc
+++ b/net/instaweb/rewriter/image_rewrite_filter.cc
@@ -2143,7 +2143,7 @@ void ImageRewriteFilter::StartElementImpl(HtmlElement* element) {
     if (srcset != nullptr) {
       BeginRewriteSrcSet(element, srcset);
     }
-	  if (datasrcset != nullptr) {
+    if (datasrcset != nullptr) {
       srcset = datasrcset ;
       BeginRewriteSrcSet(element, srcset);
     }
--- a/net/instaweb/rewriter/image_rewrite_filter.cc
+++ b/net/instaweb/rewriter/image_rewrite_filter.cc
@@ -2138,8 +2138,7 @@ void ImageRewriteFilter::StartElementImpl(HtmlElement* element) {
     BeginRewriteImageUrl(element, attributes[i].url);
    
   
-    HtmlElement::Attribute* srcset[0] = element->FindAttribute(HtmlName::kSrcset);
-    srcset[1] = element->FindAttribute(HtmlName::kDataSrcset);
+    HtmlElement::Attribute* srcset[] {element->FindAttribute(HtmlName::kSrcset),element->FindAttribute(HtmlName::kDataSrcset)};
     for (int i = 0; i == 1; i++){
        if (srcset[i] != nullptr){
 	       BeginRewriteSrcSet(element, srcset[i]);
--- a/net/instaweb/rewriter/image_rewrite_filter.cc
+++ b/net/instaweb/rewriter/image_rewrite_filter.cc
@@ -2138,7 +2138,7 @@ void ImageRewriteFilter::StartElementImpl(HtmlElement* element) {
     BeginRewriteImageUrl(element, attributes[i].url);
    
   
-    HtmlElement::Attribute* srcset[] {element->FindAttribute(HtmlName::kSrcset),element->FindAttribute(HtmlName::kDataSrcset)};
+    HtmlElement::Attribute* srcset[] = {element->FindAttribute(HtmlName::kSrcset),element->FindAttribute(HtmlName::kDataSrcset)};
     for (int i = 0; i == 1; i++){
        if (srcset[i] != nullptr){
 	       BeginRewriteSrcSet(element, srcset[i]);
--- a/net/instaweb/rewriter/image_rewrite_filter.cc
+++ b/net/instaweb/rewriter/image_rewrite_filter.cc
@@ -2138,11 +2138,13 @@ void ImageRewriteFilter::StartElementImpl(HtmlElement* element) {
     BeginRewriteImageUrl(element, attributes[i].url);
    
   
-    HtmlElement::Attribute* srcset = element->FindAttribute(HtmlName::kSrcset);
-    HtmlElement::Attribute* datasrcset = element->FindAttribute(HtmlName::kDataSrcset);
-    ( srcset == nullptr)? : BeginRewriteSrcSet(element, srcset);
-    ( datasrcset == nullptr)? : BeginRewriteSrcSet(element, datasrcset);
-
+    HtmlElement::Attribute* srcset[0] = element->FindAttribute(HtmlName::kSrcset);
+    HtmlElement::Attribute* srcset[1] = element->FindAttribute(HtmlName::kDataSrcset);
+    for (int i = 0; i == 1; i++){
+       if (srcset[i] != nullptr){
+	       BeginRewriteSrcSet(element, srcset[i]);
+        }
+    }
   }
 }
 
--- a/net/instaweb/rewriter/image_rewrite_filter.cc
+++ b/net/instaweb/rewriter/image_rewrite_filter.cc
@@ -2139,14 +2139,9 @@ void ImageRewriteFilter::StartElementImpl(HtmlElement* element) {
    
   
     HtmlElement::Attribute* srcset = element->FindAttribute(HtmlName::kSrcset);
-	  HtmlElement::Attribute* datasrcset = element->FindAttribute(HtmlName::kDataSrcset);
-    if (srcset != nullptr) {
-      BeginRewriteSrcSet(element, srcset);
-    }
-    if (datasrcset != nullptr) {
-      srcset = datasrcset ;
-      BeginRewriteSrcSet(element, srcset);
-    }
+    HtmlElement::Attribute* datasrcset = element->FindAttribute(HtmlName::kDataSrcset);
+    ( srcset == nullptr)? : BeginRewriteSrcSet(element, srcset);
+    ( datasrcset == nullptr)? : BeginRewriteSrcSet(element, datasrcset);
 
   }
 }
--- a/net/instaweb/rewriter/image_rewrite_filter_test.cc
+++ b/net/instaweb/rewriter/image_rewrite_filter_test.cc
@@ -1342,6 +1342,35 @@ TEST_F(ImageRewriteTest, ImgSrcSet) {
       "<img src=\"xa.png.pagespeed.ic.0.png\" "
       "srcset=\"xa.png.pagespeed.ic.0.png 1x, xb.png.pagespeed.ic.0.png 2x\">");
 }
+  
+TEST_F(ImageRewriteTest, ImgDataSrcSet) {
+  AddFileToMockFetcher("a.png", kBikePngFile, kContentTypePng, 100);
+  AddFileToMockFetcher("b.png", kCuppaPngFile, kContentTypePng, 100);
+
+  options()->EnableFilter(RewriteOptions::kRecompressPng);
+  rewrite_driver()->AddFilters();
+
+  ValidateExpected(
+      "data-srcset",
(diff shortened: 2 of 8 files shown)


jmarantz (commented):
This is very cool! We should definitely add a unit-test.

I think it would also be really good add an integration-test and would be happy to advise you on how to do so. Please start by looking at https://github.com/apache/incubator-pagespeed-mod/blob/master/pagespeed/automatic/system_test.sh which is a driver for the existing system-tests.

Then you might be able to modify one of the tests it calls: https://github.com/apache/incubator-pagespeed-mod/blob/master/pagespeed/automatic/system_tests/responsive_images.sh to cover the case you've added handling for.

Thanks!

net/instaweb/rewriter/image_rewrite_filter.cc
```diff
@@ -2136,13 +2136,18 @@ void ImageRewriteFilter::StartElementImpl(HtmlElement* element) {
...
-
-  if (element->keyword() == HtmlName::kImg) {
+   
+  
     HtmlElement::Attribute* srcset = element->FindAttribute(HtmlName::kSrcset);
+	  HtmlElement::Attribute* datasrcset = element->FindAttribute(HtmlName::kDataSrcset);
     if (srcset != nullptr) {
       BeginRewriteSrcSet(element, srcset);
     }
+	  if (datasrcset != nullptr) {
```
oschaaf: @Lofesa tiny nit on formatting here. I'd love to help you with an easier flow to edit these files; for example vscode makes it easier to use git, and also is able to automatically fix formatting issues for you. Please ping me on Slack if you want me to help you get set with that.

Lofesa: Changed

Lofesa:
Hi @jmarantz 
Added install/mod_pagespeed_test/image_rewriting/data-srcset.html 
 In this https://github.com/apache/incubator-pagespeed-mod/blob/master/pagespeed/automatic/system_test.sh added run_test rewrite_images_datasrcset
And in this one https://github.com/apache/incubator-pagespeed-mod/blob/master/pagespeed/automatic/system_tests/responsive_images.sh to cover  added a new: test start_test rewrite_images_datasrcset can rewrite data-srcset itself

Hope this work.

For the amp-img need to put this UrlValuedAttribute amp-img src image; in config file, but can´t find the file to put it.

jmarantz:
Lofesa, are you able to run the integration tests locally? WDYT of adding a unit-test as well?

I noticed also that you have a test for "data-srcset" but I don't see a config update to look for that attribute (e.g. with UrlValuedAttribute) nor do I see that attribute as an html standard on a web search.  It does look like it's used as an ad-hoc placeholder in some JS libraries, but I don't know if we should hard-code for that. It probably doesn't do any harm, but I just wanted to bring that up. Thoughts?

Lofesa:
Hi @jmarantz
> 
> 
> Lofesa, are you able to run the integration tests locally? WDYT of adding a unit-test as well?

I have it run in production [here](https://intersindicalrm.org/).
You can see data-srcset and amp-img srcset got rewrited. (the amp version is url/?amp).

> 
> I noticed also that you have a test for "data-srcset" but I don't see a config update to look for that attribute (e.g. with UrlValuedAttribute) nor do I see that attribute as an html standard on a web search. It does look like it's used as an ad-hoc placeholder in some JS libraries, but I don't know if we should hard-code for that. It probably doesn't do any harm, but I just wanted to bring that up. Thoughts?

My approach to this is: If you have a image element ( such as UrlValuedAttribute amp-img src image) this element maybe (or not) to have a srcset or/and a data-srcset, if it have it must be rewrittten.
I think that data-srcset is a standard "de facto", some JS libraries and other tools use it to lazyload images , in fact pagespeed uses other attribute to lazyload images that is less ussed. Why don´t help people that use data-srcset to get their images optimized?
With a long look, all this should be reconsidered, as soon browsers will use the attribute loading = "lazy" to lazyload images (In fact the lazysizes library has now an option to don´t fire if the attribute is set and the brownser can lazyload images)
Then the lazyload_images filter (as well all these JS libraries and tools) will be unnecessary, limiting itself to adding said attribute to the images if it is not added by the default by  CMS or other web development tools.

P.S. Sorry for my bad english.

jmarantz:
RE "I ran it in production" -- awesome! I believe it works. However I still hope we can maintain good software engineering discipline and have unit and integration tests that we validate are working as well, so that we ensure it stays working across future changes.

I think it's possible our unit-tests are run in CI but maybe not our integration tests,. @oschaaf can you remind me? It will also benefit you a lot to be able to run them locally for iteration.

RE "data-srcset" is defacto standard; I'm OK with that, especially as it's unlikely for it to be present in html for some other purpose. @oschaaf WDYT?

RE this whole flow being obsolete in favor of browsers directly supporting lazy...what do you recommend with respect to this PR?

Lofesa:
Hi @jmarantz 

Well.... I'm not a developer but I learn, more or less. I do not know if I would like to develop a unit test but everything is to start. Where to start?

RE:but maybe not our integration tests 
I think yes as far as these messages are in the log: 
TEST: rewrite_images_datasrcset can rewrite data-srcset itself

RE:what do you recommend with respect to this PR?
I think this PR should be merged because until the new functionality is present in all, or most, browsers will spend a lot of time and while people will continue to use data-srcset.
But preemptively you have to go thinking about what happens with the filter lazyload_images and their implications.

P.S: Sorry for my bad english.

jmarantz:
No worries -- let's start slow, and I really appreciate you pushing the product forward in all the ways you do! Happy to help.

1st: can you compile and run locally to test at all? There are instructions on setting up a development environment and running unit tests here:  https://www.modpagespeed.com/doc/build_mod_pagespeed_from_source

In general, when modifying a source file foo.cc, there'll be a unit-test foo_test.cc in the same directory. So for this change it would be https://github.com/apache/incubator-pagespeed-mod/blob/master/net/instaweb/rewriter/image_rewrite_filter_test.cc . That file is a little big and scary but you don't need to know all of it as you are only replicating similar functionality to what already exists, you can similarly replicate the test, which is ImageRewriteTest.ImgSrcSet on line https://github.com/apache/incubator-pagespeed-mod/blob/1a64c0e356f7969dff748ece6b8d11b5aa123d48/net/instaweb/rewriter/image_rewrite_filter_test.cc#L1332

Just take that test-block and copy it to a new one, change the attribute name and expected result, and you should be good to go.

The hardest part by far of this will be setting up your dev environment to run the tests. Once you have that set up I think this will be fairly easy to add a unit-test for.

oschaaf:
@jmarantz re: testing

CI runs [build_release.sh](https://github.com/apache/incubator-pagespeed-mod/blob/master/install/build_release.sh):
- This runs `pagespeed_automatic_test` and `mod_pagespeed_test` in various configurations.
- Then it builds the mod_pagespeed package, installs it, and runs the system tests in multiple configurations

RE: data-srcset
I'm +1 on as-is regarding `data-srcset`. 

@Lofesa this LGTM, but I think we can condense/simplify the code a bit. That would be great to include before merging this. Let's sync up on Slack to get that sorted.

Lofesa:
Hi @jmarantz 
I run pagespeed_automatic_test locally and only get one test fail, but this test will fail in  a plain new cloned repo:
`[ FAILED  ] InprocessShm/SharedMemStatisticsTestTemplate/0.TestHistogramRender, where TypeParam = net_instaweb::(anonymous namespace)::InProcessSharedMemEnv`

All test i have added are ok:
```
[----------] 141 tests from ImageRewriteTest
[ RUN      ] ImageRewriteTest.ImgTag
[       OK ] ImageRewriteTest.ImgTag (73 ms)
[ RUN      ] ImageRewriteTest.ImgSrcSet
[       OK ] ImageRewriteTest.ImgSrcSet (17 ms)
[ RUN      ] ImageRewriteTest.ImgDataSrcSet
[       OK ] ImageRewriteTest.ImgDataSrcSet (17 ms)
[ RUN      ] ImageRewriteTest.ImgAmpSrcSet
[       OK ] ImageRewriteTest.ImgAmpSrcSet (17 ms)
```
RE: when it is serving Amp, and maybe it should rewrite amp-img
But not. The only way the amp-img is rewrited is with
`UrlValuedAttribute amp-img src image`
whitout this PR only the src is rewrited, with this PR src and srcset get writed.

Lofesa:
> @Lofesa this LGTM, but I think we can condense/simplify the code a bit. That would be great to include before merging this. Let's sync up on Slack to get that sorted.

Well.... I make this doble if cause at first they are out of the if loop that iterates trougth the attributes of a element and need to ask for both (srcset and data-srcset). Now that are in to the loop can be simplified, sure.

In php some like this can be done:

```
swicth (true) {
   case srcset != nulltr 
          BeginRewriteSrcSet(element, srcset);
  case datasrc != nullptr
          BeginRewriteSrcSet(element, datasrcset);
  }
```
I don´t know if this is possible in c++

oschaaf:
@Lofesa 

> I run pagespeed_automatic_test locally and only get one test fail, but this test will fail in a plain new cloned repo:

Are you running the tests as root?

> Well.... I make this doble if cause at first they are out of the if loop that iterates trougth the attributes of a element and need to ask for both (srcset and data-srcset). Now that are in to the loop can be simplified, sure

great!

Lofesa:
> Are you running the tests as root?

No, i have a builder user. Don´t build nothing as root.
I´ve edited my latest post with a swchit case but don´t know if these use is possible in c++

Lofesa:
@oschaaf 

Perhaps some like this?:

```
HtmlElement::Attribute* srcset[0] = element->FindAttribute(HtmlName::kSrcset);
HtmlElement::Attribute* srcset[1] = element->FindAttribute(HtmlName::kDataSrcset);

for (int i = 0; i == 1; i++){
     if (srcset[i] != nullptr){
	BeginRewriteSrcSet(element, srcset[i]);
     }
}

```
or maybe this:

```
HtmlElement::Attribute* srcset = element->FindAttribute(HtmlName::kSrcset)
HtmlElement::Attribute* datasrcset = element->FindAttribute(HtmlName::kDataSrcset);

( srcset == nullptr)? : BeginRewriteSrcSet(element, srcset);
( datasrcset == nullptr)? : BeginRewriteSrcSet(element, datasrcset);

```

net/instaweb/rewriter/image_rewrite_filter_test.cc
```diff
@@ -1342,6 +1342,35 @@ TEST_F(ImageRewriteTest, ImgSrcSet) {
       "<img src=\"xa.png.pagespeed.ic.0.png\" "
       "srcset=\"xa.png.pagespeed.ic.0.png 1x, xb.png.pagespeed.ic.0.png 2x\">");
 }
+  
+TEST_F(ImageRewriteTest, ImgDataSrcSet) {
```
oschaaf, line 1346: @Lofesa awesome, thanks for adding these unit-tests

net/instaweb/rewriter/image_rewrite_filter.cc
```diff
@@ -2136,13 +2136,18 @@ void ImageRewriteFilter::StartElementImpl(HtmlElement* element) {
     }
 
     BeginRewriteImageUrl(element, attributes[i].url);
-  }
-
-  if (element->keyword() == HtmlName::kImg) {
+   
+  
     HtmlElement::Attribute* srcset = element->FindAttribute(HtmlName::kSrcset);
+	  HtmlElement::Attribute* datasrcset = element->FindAttribute(HtmlName::kDataSrcset);
```
oschaaf: nit: Can you fix the indentation here? Sorry about dragging on about that, but we're pretty strict about style; consistency helps everyone a lot.

pagespeed/automatic/system_tests/responsive_images.sh
```diff
@@ -35,4 +35,7 @@ start_test rewrite_images can rewrite srcset itself
 URL=$TEST_ROOT/image_rewriting/srcset.html?PageSpeedFilters=+rewrite_images,+debug
 fetch_until -save $URL 'grep -c xPuzzle.*1x.*xCuppa.*2x' 1
 
+start_test rewrite_images_datasrcset can rewrite data-srcset itself
+URL=$TEST_ROOT/image_rewriting/data-srcset.html?PageSpeedFilters=+rewrite_images,+debug
+fetch_until -save $URL 'grep -c srcset.*xPuzzle.*1x.*xCuppa.*2x' 2
```
oschaaf, line 40: Should we grep for `data-srcset` instead of just `src-set`?

oschaaf (changes requested):
I misread the code earlier, we're not in the loop that iterates the attributes when we start rewrites of the `srcset`s. Sorry about the noise, that is fine to leave as-is. 
Thanks for adding the tests!

pagespeed/automatic/system_tests/responsive_images.sh
Lofesa: Well.... In the html file we have data-srcset and srcset, this second in the amp-img element.
Whit this grep, using srcset we cover both data-srcset and amp-image srcset. Is for that that we need to count 2 in the grep condition. If use data-srcset in the grep we need other fecht_until to test amp-img srcset.... That can be ok but...We need to check the same file 2 times?

oschaaf: I see, sounds good to me as-is then, thanks.

Lofesa:
> 
> 
> I misread the code earlier, we're not in the loop that iterates the attributes when we start rewrites of the `srcset`s. Sorry about the noise, that is fine to leave as-is.
> Thanks for adding the tests!

But now the rewrite of srcset and data-srcset is inside the loop and we can simplify the double if.
We can use the if .. else, the for...if or use the ternary operator.
Personally prefer the ternary operator, less lines = simply to read.

net/instaweb/rewriter/image_rewrite_filter.cc
Lofesa: I will wait until let`s decide what to use to simplify the srcset and data-srcset if statement.

Lofesa: chnaged and added an alternative to the double if

Lofesa:
I close and reopen this PR cause the build had failed when contacting to ppa ubuntu

oschaaf:
@Lofesa just tested this in my machine and it works fine there as well, awesome.
I did also try it with this modification, which worked too:

```diff
oschaaf@burst:~/code/incubator-pagespeed-mod$ git diff
diff --git a/net/instaweb/rewriter/image_rewrite_filter.cc b/net/instaweb/rewriter/image_rewrite_filter.cc
index 10bfca442..8d0fa1153 100644
--- a/net/instaweb/rewriter/image_rewrite_filter.cc
+++ b/net/instaweb/rewriter/image_rewrite_filter.cc
@@ -2134,18 +2134,15 @@ void ImageRewriteFilter::StartElementImpl(HtmlElement* element) {
         continue;
       }
     }
-
     BeginRewriteImageUrl(element, attributes[i].url);
-   
-  
-    HtmlElement::Attribute* srcset = element->FindAttribute(HtmlName::kSrcset);
-    HtmlElement::Attribute* datasrcset = element->FindAttribute(HtmlName::kDataSrcset);
-    if (srcset != nullptr) {
-           BeginRewriteSrcSet(element, srcset);
-    } else if (datasrcset != nullptr) {
-           BeginRewriteSrcSet(element, datasrcset);
-    }      
-  
+  }
+
+  HtmlElement::Attribute* srcset = element->FindAttribute(HtmlName::kSrcset);
+  HtmlElement::Attribute* datasrcset = element->FindAttribute(HtmlName::kDataSrcset);
+  if (srcset != nullptr) {
+    BeginRewriteSrcSet(element, srcset);
+  } else if (datasrcset != nullptr) {
+    BeginRewriteSrcSet(element, datasrcset);
   }
 }
```

That is a little bit more efficient as we won't be calling FindAttribute() in a loop. 
It may be hypothetical, but I think this may also fix an issue if there are no image-valued attributes (because we will not enter the loop in that case)
WDYT?

Lofesa:
Hi @oschaaf
I think the loop is needed or we need a double if w/o the else if. Why? Think in an img element, this may have this structure:
Reformas Laborales

This element have a placeholder srcset set to a data uri (but maybe a url to a 1x1 px image) and a data-srcset with the real images. The `if else if` found the srcset placeholder and send it to BeginRewriteSrcSet but don´t ask for more image valued attributes so the data-srcset can´t be rewrited then we need 2 `if`, 1 for the srcset and 1 more for the data-srcset.

If the ineficiency is due we call FindAttribute(), maybe we ask for some like 
```
if (attributes[i]->keyword() == HtmlName::kDataSrcset) {
    BeginRewriteSrcSet(element, datasrcset);
} else if (attributes[i]->keyword() == HtmlName::kSrcset) {
   BeginRewriteSrcSet(element, srcset);
}
```

if this is the rigth syntax so we don´t call a external class.

oschaaf:
@Lofesa you are right, we shouldn't have `else if` when we're outside of the loop. So I think we should move the code out of the loop, and remove the `else`. Ideally we also test the input you described above so we prove out we got it right. Staying out of that loop make the code more efficient, so I feel it's worth getting that right if possible!

Lofesa:
@oschaaf  Well... out of the loop we need 2 `if` and we need to not ask for and img element as in the original image_rewrite_filter.cc file (line 2141) cause not only img elements can have srcset/data-srcset.
Maybe we change the line 2141 w/ some like this:
`if (element->keyword() == semantic_type::kImage) {`
before asking for srcset/data-srcset attributes?

oschaaf (approved):
We had a bit of a discussion on Slack about this, and the plan is that we'll merge this as-is, and follow up with another PR to introduce a `srcset` semantic type and clean this up a bit.
Thanks @Lofesa for this contribution (and also for being persistent!).

PR #1900, opened 2019-06-17
Standard configuration for Varnish: disable html cache is optional

csandanov:
Normally should be disabled only when intermediary caching proxy/CDN used https://github.com/apache/incubator-pagespeed-ngx/issues/1655. 

Otherwise, the section does not make sense: it starts from how to make HTML cachable but adds a no-cache option in Varnish anyway.

### DIFF
--- a/html/doc/downstream-caching.html
+++ b/html/doc/downstream-caching.html
@@ -105,15 +105,15 @@ sub vcl_recv {
   }
 }
 
-# Mark HTML as uncacheable.  If we can't send them purge requests they can't
+# Optionally when additional intermediary caching proxy/CDN used: mark HTML as uncacheable.  If we can't send them purge requests they can't
 # cache our html.
-sub vcl_fetch {
-   if (beresp.http.Content-Type ~ "text/html") {
-     remove beresp.http.Cache-Control;
-     set beresp.http.Cache-Control = "no-cache, max-age=0";
-   }
-   return (deliver);
-}
+#sub vcl_fetch {
+#   if (beresp.http.Content-Type ~ "text/html") {
+#     remove beresp.http.Cache-Control;
+#     set beresp.http.Cache-Control = "no-cache, max-age=0";
+#   }
+#   return (deliver);
+#}
 
 sub vcl_hit {
   # Make purging happen in response to a PURGE request.  This happens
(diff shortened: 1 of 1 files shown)


html/doc/downstream-caching.html
```diff
@@ -105,15 +105,15 @@ <h2 id="standard">Standard Configuration</h2>
   }
 }
 
-# Mark HTML as uncacheable.  If we can't send them purge requests they can't
+# Optionally when additional intermediary caching proxy/CDN used: mark HTML as uncacheable.  If we can't send them purge requests they can't
```
oschaaf: But .. we also do not want browsers to cache the html, no?

oschaaf (changes requested):
So thinking more about this, visitors may also be routed through transparant/forward proxies and not even know about it. We do not want those intermediaries to cache the html either right? 

I think the safe default is to not allow any caching of html downstream. If a site-owner is really 100% sure none of the sites visitors aren't routed through any intermediaries then not overriding these controls like we do now may be OK. But I think that is an uncommon scenario? 

Alternatively, I do wonder if there would be any harm in allowing `private` cacheability of html in some cases. Semantically the response should be OK, so then browser could re-use it (even if unoptimized or partially optimized) if the webserver indicated that was possible initially.

csandanov:
I'm new to pagespeed and maybe I just don't get something but I was reading this article and as I understand it is all about how to have cache while still using pagespeed? This is what I don't understand, the article starts with:

> By default PageSpeed serves HTML files with Cache-Control: no-cache, max-age=0 so that changes to the HTML and its resources are sent fresh on each request. **The HTML can be cached,** however, if you:

So, I have ngx_pagespeed and as I understand when I enable it, it adds no-cache header to all text/html. And there's a list of actions that I should follow to have the caching of text/html. Now, here's what I don't get – how html caching will be enabled? I have two options in mind:
1. I should add additional configuration on varnish to specifically ignore no-cache header for text/html. This option sounds weird to me
2. Just setting `pagespeed DownstreamCachePurgeLocationPrefix` on Nginx will tell pagespeed to no longer set no-cache header for text/html. But if it's so, then why would we add no-cache header on Varnish? If we do we won't achieve what we wanted in the first place – to cache HTML despite pagespeed

oschaaf:
So.. Varnish is allowed to cache html emitted by the pagespeed-enabled webserver with downstream caching enabled. But no-one else is allowed to cache it. This is by design. 

VAs such, Varnish should override the response headers just before it emits them to indicate that no-one is allowed to cache the html. This prevents other intermediaries and browsers from caching it. 

I do wonder if it would make sense to slightly widen this and allow browers to cache the html as well.

csandanov:
> just before it emits them to indicate that no-one is allowed to cache the html

then it should be in `vcl_delivery` instead of `vcl_backend_response` because cacheability on Varnish decided after backend response.

Setting cache to `private` sounds reasonable if all we want is to forbid the page from caching on reverse proxies.

csandanov:
> Varnish is allowed to cache html emitted by the pagespeed-enabled webserver

I guess you mean that if you have no other configuration on Varnish other than what's specified here it will still cache. And that is true but usually, people don't use Varnish without any config and when there's a config it usually is mindful of `no-cache; max-age=0` headers in `vcl_backend_response` step, and in this case, you should somehow recognize whether it's just a pagespeed header that can be ignored (due to enabled downstream caching) or it's a header coming from a backend that cannot be ignored.

oschaaf:
So thinking about this, what would be helpful is adding an actual production-like config to the docs. Would you be able to constribute that? Following the discussion here I'm inclined to think the current bare-minimum example we have now also serves a purpose for those that add Varnish with no other purpose then getting the downstream caching feature working.

oschaaf:
Closing this due to inactivity. Feel free to re-open.

PR #1904, opened 2019-07-03
Add third_party/base64 & third_party/brotli

oschaaf:


ashishk-1 (approved):
looks good

oschaaf:
This may need a little bit of work later on when we want to use these dependencies to get the include paths right when the dependencies are put to use. But I feel this is a first good step

PR #1906, opened 2019-07-05
Adding zlib to bazel build

ashishk-1:


bazel/repositories.bzl
```diff
@@ -16,3 +18,10 @@ def mod_pagespeed_dependencies():
         url = "https://github.com/google/brotli/archive/%s.tar.gz" % BROTLI_COMMIT,
     )
 
+    http_archive(
+        name = "zlib",
```
oschaaf, line 23: I think bazel has a convention to name these, but I don't know from the top of my head. Renaming this also disambiguates with the `cc_library` rule below. Maybe add a TODO so we can do this later?

bazel/zlib.bzl
```diff
@@ -0,0 +1,35 @@
...
+        "src/inflate.h",
+        "src/inftrees.h",
+        "src/trees.h",
+        "src/zconf.h",
+        "src/zlib.h",
+        "src/zutil.h",
+        "src/gzguts.h",
+    ]),
+    copts = [
+        "-w",
```
oschaaf, line 30: consider filing an issue in the bazel milestone to track 
- investigating if we need/want these options?
- revisit this to see if we need to consider the `USE_SYSTEM_ZLIB` flag

oschaaf (changes requested):
Thanks, mostly looks good, a couple of requests that I'd like to see addressed and then we can merge this.

bazel/zlib.bzl
```diff
@@ -0,0 +1,35 @@
+zlib_build_rule = """
+cc_library(
+    name = "zlib",
+    srcs = glob([
```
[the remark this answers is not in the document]
ashishk-1: removed glob from sources and headers. Now artifacts are getting shown up in bazel-bin/external

bazel/repositories.bzl
```diff
@@ -16,3 +18,10 @@ def mod_pagespeed_dependencies():
...
+        strip_prefix = "zlib-%s" % ZLIB_COMMIT,
+        url = "https://github.com/madler/zlib/archive/%s.tar.gz" % ZLIB_COMMIT,
```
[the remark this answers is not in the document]
ashishk-1: added sha256 checksum in http_archive

bazel/repositories.bzl
ashishk-1: added TODO to rename library as per bazel naming conventions.
As of now, I have used name **"zlib"** since building cc_library directly generates **libzlib.a** library (file with same name getting generated from gyp build of third party libraries)

bazel/zlib.bzl
ashishk-1: Filed issue in bazel milestone
#1907

bazel/repositories.bzl
oschaaf: I think the `.a` name is determined below via 

```
cc_library(
    name = "zlib",
```

regardless, lgtm as-is, thanks for adding the todo

oschaaf (approved):
Great, thanks, looks good to me.

PR #1908, opened 2019-07-08
Added brotli_enc and brotli_dec library targets to third party bazel build

ashishk-1:


oschaaf:
This doesn't build because of some warnings, and we're pretty much in super-strict mode.
This change LGTM, I'll look into
- See if we can apply `--copt -Wno-implicit-fallthrough` to the build of just this depedency
- Add the bazel build to travis CI

oschaaf (approved):

PR #1912, opened 2019-07-09
adding hiredis library to bazel build

sambhajis-gdb:


bazel/hiredis.bzl
```diff
@@ -0,0 +1,23 @@
...
+    name = "hiredis",
+    srcs = [
+        "async.c",
+        "hiredis.c",
+        "net.c",
+        "read.c",
+        "sds.c",
+    ],
+    hdrs = [
+        "dict.c",
```
oschaaf, line 13: So first I was wondering: should this be declared in `srcs` instead of `hdrs`?
But after digging in, I think you made the right choice here.

oschaaf (changes requested):
The dict include is kind of weird at first glance, but then I noticed `async.c` including it, so I guess
this makes sense, we must propagate this oddity.
This looks good to me, but maybe you can add a comment about why we have the `.c` file in `hdrs` to help others
after us?

bazel/hiredis.bzl
sambhajis-gdb: added comments to bazel file

oschaaf (approved):
Thanks!

PR #1914, opened 2019-07-11
adding re2 bazel build

sambhajis-gdb:


oschaaf:
Thanks!

So looking at https://github.com/google/re2/blob/e2deff36afd0a71adbf71cb11fba71ed2a407e4f/BUILD it seems this dependency could be added as a bazel native dependency.. Trying that, it seems that that no longer builds with recent bazel versions.

So what I do wonder about is: should we investigate upgrading re2 on the master branch?
If that works, we can probably go bazel native here by depending on a more recent version, instead of rolling our own `BUILD` file and injecting that.

So, let me attempt to upgrade `re2` on master, and follow up here to let you know how that went. If that turns out to be non-trivial, this is good as-is in my opinion, but if it's easy to upgrade `re2` it may make sense to go bazel-native here.

oschaaf:
I think we can upgrade `re2`: https://github.com/apache/incubator-pagespeed-mod/pull/1915
ptal!

oschaaf:
One note: I think `re2//:benchmark` replaces `re2//:bench_util` in the new `re2` version.
`mod_pagespeed_speed_test` relies on that. I think we should file an issue on this, and move on, we'll get to this once we get around to getting `mod_pagespeed_speed_test` to build, which will probably be a little further down the road.

oschaaf:
PR for backporting the re2 upgrade to the `bazel` branch: https://github.com/apache/incubator-pagespeed-mod/pull/1916

oschaaf (approved):
Lgtm, thanks!

sambhajis-gdb:
issue file for benchmark target #1917

PR #1915, opened 2019-07-11
Upgrade our re2 dependency to release 2019-07-01

oschaaf:
Backporting this to the bazel branch will help adding `re2` as a native dependency.
The version we currently depend on no longer builds with recent bazel versions.

re2 commit: 848dfb7e1d7ba641d598cb66f81590f3999a555a
 

ashishk-1 (approved):
looks good to me

PR #1920, opened 2019-07-12
Bazel/libpng: partial port

oschaaf:
This needs revisiting later on when we want to link against the sytem
libary, but for now this might just suffice as a step closer towards
pagespeed_automatic_(test)


ashishk-1 (approved):
looks good to me

PR #1922, opened 2019-07-12
Bazel/ICU: initial stab at building ICU

oschaaf:


ashishk-1 (approved):
looks good to me

PR #1923, opened 2019-07-13
Update CI configuration

PeterDaveHello:


oschaaf (approved):
Nice! This looks great to me, the caching potentially shaves lots of minutes from the CI run duration. Ok with me merging this?

.travis.yml
```diff
@@ -1,4 +1,5 @@
 language: c++
+dist: xenial
```
oschaaf: This is kind of tricky, we use the older distro on purpose: we build releases on an ancient version for compatibility reasons. So currently we run CI on the same platform that we actually build releases with.

PeterDaveHello: Let me rollback this one!

PeterDaveHello:
The build almost made it but finally exceeded the maximum time limit for jobs when packing up cache, I'll try trigger the build a few times again, since the build time varies on Travis, once we passed it, the future builds will be faster!

PeterDaveHello:
Looks like most of the recent builds of master branch already been close to the default max build time - 50 mins, which makes some additional adjustments hard to being merged... Using force push to re-triggered seems to be a bad choice, I wonder if any maintainers here can help write an email to Travis CI asking for extending the max build time or help retry the build?

oschaaf:
@PeterDaveHello just triggered a re-run. Whom should I ask to extend the build-time? I can write an email and ask if they can help with that,  but I'm not sure who to direct that to.

PeterDaveHello:
I used to use support@. Looks like it passed now, hopefully the future builds will be faster after the build of merge commit so maybe the request won't be needed ;)

oschaaf:
Cool! Is this ready to merge?

PeterDaveHello:
Yes, I think so!

oschaaf (approved):

oschaaf:
Thanks @PeterDaveHello !

oschaaf:
So with this merged I can’t get CI on master to pass because of exceeding the max time. I wonder how that could be, the caching ought to speed things up right?

PeterDaveHello:
Both ccache and git submodule cache has their overhead, so before the first cache been established, it make sense to extended the build time a little bit, before the patch, build time is already about 48 mins, which is very close to 50 mins, and also, build time varies on Travis CI, depends on the loading of the cloud, that's why we need "retry" in this PR, to get a chance to pass it(as one of the previous builds shows the build was very close to be finished, so I think it worth a try).

oschaaf:
Just restarted it, 5th attempt :)

PeterDaveHello:
Thanks, wish to have the passed result ASAP!

oschaaf:
OK, so we've now done quite a bit of attempts, so I'm giving up on that. I've send an email to support@ to request an extension of the deadline, I'll follow up here once I've heard more

oschaaf:
They responded that they extended the job timeout, and I tried, and we're passing on master again!

PeterDaveHello:
Nice! Looks like it takes less than 53 mins, about 4~5 mins overhead to enable ccache and submodule cache, let's see if future builds get faster after this one!

PR #1924, opened 2019-07-13
Bazel/libwebp: build with bazel

oschaaf:
First stab

- Needs buildifier reformatting
- Needs revisiting for the static library build


ashishk-1 (approved):
looks good to me

PR #1927, opened 2019-07-16
[WIP] bazel/absl: pagespeed/kernel

oschaaf:
Kind of a messy state dump, putting this out there for those who want an early look.
Needs quite a bit of cleanup, probably we need to go over this and carve pieces off to land
it bit-by-bit once `pagespeed/kernel` is completely ready.

/cc @ashishk-1 @jmarantz


oschaaf:
This is now able to run tests in `/pagespeed/*` & `net/instaweb/*`, the vast majority passing. 
however:
- Tests that rely on `third_party/css parser` fail, probably because I messed up somewhere.
  This needs debugging.
- `HMAC()` segfaults somehow, some tests call that. This also needs sorting out.

PR #1928, opened 2019-07-16
prioritize critical css : avoid adding duplicate ids 

### LINKED ISSUE #1878: prioritize critical css need to remove id attributes
Hi
Using the prioritize_critial_css inlined styles with id attibutes get duplicate id´s, one when the filter restore the inlined style (if it is in the critical path) and other when the original inlined style is loaded on the onload event.

Lofesa:
This is the same as PR #1886 , closed when the branch is removed.
PR #1886  was awaiting a request for review by @jmarantz

### DIFF
--- a/net/instaweb/rewriter/critical_css_loader.js
+++ b/net/instaweb/rewriter/critical_css_loader.js
@@ -48,7 +48,7 @@ pagespeed.CriticalCssLoader.addAllStyles = function() {
     for (var v = 0; v < children.length; ++v) {
       children[v].removeAttribute('id');
     }  
-  document.body.appendChild(div);
+    document.body.appendChild(div);
   }
 };
 
--- a/net/instaweb/rewriter/critical_css_loader.js
+++ b/net/instaweb/rewriter/critical_css_loader.js
@@ -39,11 +39,16 @@ pagespeed.CriticalCssLoader.addAllStyles = function() {
 
   var elements = document.getElementsByClassName('psa_add_styles');
 
+  
   for (var i = 0, e; e = elements[i]; ++i) {
     if (e.nodeName != 'NOSCRIPT') { continue; }
     var div = document.createElement('div');
     div.innerHTML = e.textContent;
-    document.body.appendChild(div);
+    var children = div.childNodes;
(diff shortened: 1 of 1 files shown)


Lofesa:
Close and reopen to run CI again

oschaaf:
Cross linking issue this should close: https://github.com/apache/incubator-pagespeed-mod/issues/1878

oschaaf (changes requested):
@Lofesa the change makes sense to me, however, I think we also need to consider the child elements of the child elements, and so on. So we should recursively strip id attributes. Would you be up for making that change?

net/instaweb/rewriter/critical_css_loader.js
```diff
@@ -39,11 +39,16 @@ pagespeed.CriticalCssLoader.addAllStyles = function() {
...
   for (var i = 0, e; e = elements[i]; ++i) {
     if (e.nodeName != 'NOSCRIPT') { continue; }
     var div = document.createElement('div');
     div.innerHTML = e.textContent;
-    document.body.appendChild(div);
+    var children = div.childNodes;
+    for (var v = 0; v < children.length; ++v) {
+      children[v].removeAttribute('id');
+    }  
+  document.body.appendChild(div);
```
oschaaf: nit: indent

Lofesa: The change you requested is to indent the line
  `document.body.appendChild(div);` ?

oschaaf: yes, that would be great. But we may also want to recurse? So not only remove the `id` attribute from the direct children of the element, but also the children of the children, and so on?

Lofesa: But childNode don´t get all childs and sub-childs nodes?

oschaaf: Oh in that case you’re right this is OK.

oschaaf: Sorry about the noise!

oschaaf (approved):

PR #1929, opened 2019-07-16
Data srcset

Lofesa:
A new version for rewrite srcset and data-srcset. Maybe doc must be updated to add
`pagespeed AddUrlValuedAttribute amp-img (data)-srcset srcsetimage`
Related to PR #1899


jmarantz:
I'm concerned about amp-img. I feel like we might not want to rewrite URLs of such images, but only rewrite their content as part of IPRO. The reason is that the rewrites are user-agent-specific, and that amp HTML will be cached across user-agents.

Lofesa:
> I'm concerned about amp-img. I feel like we might not want to rewrite URLs of such images, but >only rewrite their content as part of IPRO. The reason is that the rewrites are user-agent-specific, >and that amp HTML will be cached across user-agents.
What do you mean with "..that amp HTML will be cached across user-agents" ?
If the page is served from the origin server, the images are served acordly to the browser user agent, if served from cdn.ampproject.org, the images are served w/ the optimizations for their UA.
Maybe we exclude it from pagespeed if the referal is cdn.ampproject.org.
Some like this in nginx:
```
map $http_referer $bad_referer {
    default                  0;
    "~* .*cdn.ampproject.org.*"       1;
}
if ($bad_referer) {
    pagespeed off;
}
```
EDIT: More on this. I think that amp pages for cdn.ampproject.org must be served with PageSpeed=off, this cdn uses a custom version of pagespeed, I can see this header i it: 

x-page-speed | 0.9.10.99-9999

EDIT2: Seem that pagespeed off; is not allowed in a if in nginx
EDIT3: Found this:
```
map $http_referer $bad_referer {
    default                  0;
    "~* .*cdn.ampproject.org.*"       1;
}
if ($bad_referer) {
    set $args &args&PageSpeed=off;
}
```

net/instaweb/rewriter/url_left_trim_filter.cc
```diff
@@ -65,6 +65,7 @@ void UrlLeftTrimFilter::StartElementImpl(HtmlElement* element) {
     resource_tag_scanner::ScanElement(
         element, driver()->options(), &attributes);
     for (int i = 0, n = attributes.size(); i < n; ++i) {
+      if (attributes(i).category == semantic_type::kSrcSetImage) {continue;}
```
jmarantz: put the `continue;` on a separate link, which makes it easier to read and debug.

net/instaweb/rewriter/cache_extender.cc
```diff
@@ -250,6 +250,24 @@ void CacheExtender::StartElementImpl(HtmlElement* element) {
         driver()->InitiateRewrite(context);
       }
     }
+    HtmlElement::Attribute* datasrcset = element->FindAttribute(HtmlName::kDataSrcset);
+    if (datasrcset != nullptr) {
```
jmarantz: Can you call this local variable `data_srcset`? I think that would make it easier to read.

jmarantz (commented):
Thanks for digging into this further! 

What do you think about adding some unit tests?

net/instaweb/rewriter/url_left_trim_filter.cc
Lofesa: Sure

net/instaweb/rewriter/cache_extender.cc
Lofesa: Sure

Lofesa:
@jmarantz 
> 
> 
> Thanks for digging into this further!
> 
> What do you think about adding some unit tests?

About the amp-img:
cdn.ampproject.org fecht resources with the Google bot for mobil UA.

Unit test for srcset, data-srcset and amp-img src and srcset are added in a previous PR #1899 so I think in this PR are not needed or are you talking about others unit test?

Lofesa:
In the file `net/instaweb/rewriter/cache_extender.cc` I have duplicated the structure that start at line 233
```
if (element->keyword() == HtmlName::kImg &&
      driver()->MayCacheExtendImages()) {
    HtmlElement::Attribute* srcset = element->FindAttribute(HtmlName::kSrcset);
    if (srcset != nullptr) {
```
to catch the case attribute is data-srcset.

This maybe simplyfied by:

```
if (attribute(i).category == semantic_type::kSrcSetImage &&
       driver()->MayCacheExtendImages()) {
     HtmlElement::Attribute* data_srcset = attribute[i].url;
     SrcSetSlotCollectionPtr slot_collection(
          driver()->GetSrcSetSlotCollection(this, element, data_srcset));
```
and don´t have these duplicated for srcset and data-srcset, but I´m unsure.
Can some one take a look?

.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,18 @@
+{
```
oschaaf: This file doesn't belong here I guess, can you back it out?

oschaaf (commented):
@Lofesa my apologies in the huge delay here; I'm very much snowed under at the moment. What I suggest doing is:
- write up a short doc that describes, high level, what this should do, preferably with examples of input and output. https://github.com/apache/incubator-pagespeed-mod/wiki/Design-Documents would be a great place for that.
- subsequently I can contribute unit tests based on that.
- however, i think we should wait for a little bit and take this to the bazel branch. I'll think I can get us coverage reports there, which greatly helps reviewers assess if testing is sufficient.

Lofesa:
@oschaaf 
Some unit test where set in the PR #1899 . Amp-img and data-srcset, if I remember, so maybe only a unit test for UrlValuedAttribute is needed.

For the doc... some like this?

[PR-Data-srcset.pdf](https://github.com/apache/incubator-pagespeed-mod/files/3901157/PR-Data-srcset.pdf)

oschaaf:
Closing this in an effort to clean up stale pull requests (but feel free to re-open if you want to pick this up again).

PR #1930, opened 2019-07-17
adding giflib bazel build

sambhajis-gdb:


oschaaf (approved):
Thanks! I do wonder about the `giflib_gen_arch_root` stuff in the gyp we're porting here, but we'll figure that out once we need to consume this lib (if we need to).
LGTM.

PR #1935, opened 2019-07-28
[WIP] Bazel build system

oschaaf:
Early PR with a much leaner port to the bazel build system, for those who are interested.
There are still some string changes in here, but those are required for us to be able to translate buffers when communicating with `third_party/css_parser`.

- This builds with close-to-vanilla recent chromium url/base dependency.
- Most string `absl` related changes are left out and taken care of in a shim `StringPiece` class. 
- All unit tests passing in `net/..` and `pagespeed/...` except for:
  - tests that expect closure-compiled js, which isn't incorporated in the build flow yet
  - HMAC/sha1 tests segfaulted, which is worked around & passing, but still needs root-cause analysis.


ashishk-1 (approved):
Looks good

pagespeed/kernel/base/string_util.h
```diff
@@ -17,48 +17,182 @@
...
 
-using base::StringAppendF;
-using base::StringAppendV;
-using base::SStringPrintf;
-using base::StringPiece;
-using base::StringPrintf;
+static const int32 kint32max = 0x7FFFFFFF;
+static const int32 kint32min = -kint32max - 1;
+
+using namespace absl;
```
jmarantz: I would recommend avoiding this; I'm scared of what we might grow to be dependent on in the presence of this. Instead just inherit from "public absl::string_view" below.

pagespeed/kernel/base/string_util.h
```diff
@@ -17,48 +17,182 @@
...
+public:
+  StringPiece(const absl::string_view& s) : absl::string_view(s) {}
+  StringPiece(const GoogleString& s) : absl::string_view(s.data(), s.size()) {}
+
+  using string_view::string_view;
+  void CopyToString(GoogleString* dest) const {
+    *dest = std::string(*this);
+  }
+  void AppendToString(GoogleString* dest) const {
+    (*dest).append(std::string(*this));
```
jmarantz: this temp can be avoided.

pagespeed/kernel/base/string_util.h
```diff
@@ -17,48 +17,182 @@
...
+  bool starts_with(StringPiece prefix) const {
+    return absl::StartsWith(*this, prefix);
+  }
+
+  bool ends_with(StringPiece postfix) const {
+    return absl::EndsWith(*this, postfix);
+  }
+
+  void set(StringPiece newvalue, uint32_t size) {
+    absl::string_view tmp(newvalue.data(), size);
```
jmarantz, line 81: can assign to *this.

pagespeed/kernel/base/string_util.h
oschaaf: Fixed, this is now constrained to

```
using absl::StrCat;
using absl::StrAppend;
```

oschaaf: done

pagespeed/kernel/base/string_util.h
oschaaf: done

pagespeed/kernel/base/string_util.h
oschaaf: done

base/base.gyp
```diff
@@ -1,63 +0,0 @@
-# Copyright (c) 2009 The Chromium Authors. All rights reserved.
-# Use of this source code is governed by a BSD-style license that can be
```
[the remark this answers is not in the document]
oschaaf: While we're not confident that the `bazel` branch is on-par, changes that land on `master` need to be cherry-picked here. I wasn't thinking about merging this anytime soon yet, it's still a WIP. Running two branches seems less work compared to attempting to do it on a single branch (though not impossible of course), because of the current grpc version on master being non-trivial to port. There's some bloat from the updated/leaned up chromium dependencies; it would be a bit more work to get the bazel branch working with the current version instead of updating the dependency right away.

oschaaf:
Closing this will, open a new tracking PR based on the `bazel` branch

PR #1939, opened 2019-08-22
Building PageSpeed with bazel & absl

oschaaf:
Early PR to make it easy for those who are interested to track work-in-progress on
building pagespeed with bazel & absl.

State:

- There are still some string changes in here, but those are required for us to be able to translate buffers when communicating with third_party/css_parser.
- This builds with close-to-vanilla recent chromium url/base dependency.
- Most string absl related changes are left out and taken care of in a shim StringPiece class.
- All unit tests passing.
analysis.
- Contains an experimental work-in-progress for an Envoy filter

/cc @ashishk-1 @jmarantz


oschaaf:
Merged the Envoy filter skeleton in here, because maintaining two branches is becoming a drag.
Disabled the build/test for that in Travis

sgammon:
this is awesome and you've got a tester right here

sgammon:
early feedback on the early PR. failed to build on macOS and linux (GCP Cloud Shell instance).

what follows is the error on both platforms, in case it might be useful:

**linux:**
```
envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto: warning: Import validate/validate.proto but not used.
envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto: warning: Import udpa/annotations/status.proto but not used.
envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto: warning: Import udpa/annotations/versioning.proto but not used.
INFO: From ProtoGenValidateCcGenerate external/envoy_api/envoy/config/transport_socket/raw_buffer/v2/raw_buffer.pb.h:
envoy/config/transport_socket/raw_buffer/v2/raw_buffer.proto: warning: Import udpa/annotations/status.proto but not used.
envoy/config/transport_socket/raw_buffer/v2/raw_buffer.proto: warning: Import udpa/annotations/migrate.proto but not used.
ERROR: /home/sam_g/.cache/bazel/_bazel_sam_g/442aefc8b303119970adb9b4d641e482/external/icu/BUILD.bazel:29:1: C++ compilation of rule '@icu//:icuuc' failed (Exit 1) gcc failed: error executing command /usr/bin/gcc -U_FORTIFY_SOURCE -fstack-protector -Wall -Wunused-but-set-parameter -Wno-free-nonheap-object -fno-omit-frame-pointer '-std=c++0x' -MD -MF ... (remaining 32 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
gcc: error: unrecognized command line option '-fno-addrsig'; did you mean '-Wno-address'?
Target //pagespeed/envoy:envoy failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 890.303s, Critical Path: 61.68s
INFO: 801 processes: 801 linux-sandbox.
FAILED: Build did NOT complete successfully
```

**macOS:**
```
INFO: From Compiling external/icu/source/common/locmap.c [for host]:
external/icu/source/common/locmap.c:165:1: warning: unused variable 'locmap_ber' [-Wunused-const-variable]
ILCID_POSIX_SUBTABLE(ber) {
^
external/icu/source/common/locmap.c:99:32: note: expanded from macro 'ILCID_POSIX_SUBTABLE'
static const ILcidPosixElement locmap_ ## id [] =
                               ^
<scratch space>:41:1: note: expanded from here
locmap_ber
^
1 warning generated.
INFO: From Compiling external/icu/source/common/locdispnames.cpp [for host]:
external/icu/source/common/locdispnames.cpp:284:19: warning: unused variable '_kRootName' [-Wunused-const-variable]
static const char _kRootName[]        = "root";
                  ^
1 warning generated.
INFO: From Compiling external/icu/source/common/normlzr.cpp [for host]:
external/icu/source/common/normlzr.cpp:67:20: warning: unused variable '_NUL' [-Wunused-const-variable]
static const UChar _NUL=0;
                   ^
1 warning generated.
ERROR: /private/var/tmp/_bazel_sam.g/54a21370b297491b372e1fb2356c09a0/external/icu/BUILD.bazel:29:1: C++ compilation of rule '@icu//:icuuc' failed (Exit 1) wrapped_clang failed: error executing command external/local_config_cc/wrapped_clang '-D_FORTIFY_SOURCE=1' -fstack-protector -fcolor-diagnostics -Wall -Wthread-safety -Wself-assign -fno-omit-frame-pointer -g0 -O2 '-D_FORTIFY_SOURCE=1' -DNDEBUG ... (remaining 30 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
external/icu/source/common/putil.c:622:12: error: use of undeclared identifier '__timezone'; did you mean 'timezone'?
    return U_TIMEZONE;
           ^~~~~~~~~~
           timezone
external/icu/genfiles/arch/linux/common/include/unicode/platform.h:292:25: note: expanded from macro 'U_TIMEZONE'
#define U_TIMEZONE      __timezone
                        ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/usr/include/time.h:103:13: note: 'timezone' declared here
extern long timezone __DARWIN_ALIAS(timezone);
```

oschaaf:
@sgammon great to hear! I think this is currently broken for gcc, but clang ought to work. I need to look into that.  Also I need to update the PR description. This does actually produce usable artifacts these days, including mod_pagespeed.so :-) the Envoy filter still is highly experimental still though, and only supports IPRO mode.

sgammon:
@oschaaf cool! i'll try it with `clang` (assuming on Linux, due to the `sys/sendfile.h` include). if you have any other build env recommendations, let me know. i'm on macOS locally and GKE/COS in prod.

currently, we fly pagespeed in `nginx`, but behind `envoy` to facilitate gRPC-web routing, like so:

```
                   ---- nginx
L7 GCLB > Envoy ---|     |
                   ---- gRPC-Java
```

the Nginx installation here is actually two processes: one Nginx process configured with Pagespeed for web optimization, and one Nginx process configured as an [ESPv1](https://github.com/cloudendpoints/esp) instance, to do JSON transcoding so our gRPC API can be used with browsers that don't support `fetch` (and therefore cannot support [`gRPC-web`](https://github.com/grpc/grpc-web)).

i mention this context because, with the new [ESPv2](https://github.com/GoogleCloudPlatform/esp-v2) beta, they're also moving to Envoy. that means, with _this_ PR, a full-Google-native set up on GCP can finally just look like:

`L7 GCLB` > `Envoy` > `gRPC-Java`

but with creative routing/filtering to enable pagespeed vs. gRPC on different endpoints. and all of it can soon be built with Bazel, which we are also big adopters of. obviously that would be a huge win for us, and people on our stack, so you have our attention for sure! :)

oschaaf:
If all else fails, maybe the [CI config](https://github.com/apache/incubator-pagespeed-mod/blob/bazel/.travis.yml) can be helpful to figure out what's needed to bully the build into succeeding. warning: it can take quite a while. That's why the Envoy build is excluded over there: it takes more time then travis allows us.
 
I think the plan you lay out would be viable, and NGINX could be taken out of the equation if it's only used as a vehicle to run pagespeed. But the Envoy port in here may need polish to make it comply with the routing you'd need, depending on how it needs to be configured, and it's still early stage / has no production mileage.

High level, the rationale of the work here is project modernization, as well as creating (lots of) opportunities to lean up the project by piggy backing on Envoy/absl/bazel for a few things.
 
Second, in a next step, the idea is to make the primary modus operandi running PageSpeed (v2) as a sideca for non-Envoy servers, either by documenting the required routing configuration or by providing a very thin module to set that up automatically. 
As Envoy is the vehicle of choice to facilitate running as a sidecar, no sidecar would be needed, as it will be able to run pagespeed directly with this extension compiled in.

oschaaf:
@sgammon Oh -- travis aside, the Envoy build requirements should be satisfied:
https://www.envoyproxy.io/docs/envoy/latest/install/building

base/debug/leak_annotations.h
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) 2011 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef BASE_DEBUG_LEAK_ANNOTATIONS_H_
+#define BASE_DEBUG_LEAK_ANNOTATIONS_H_
+
+#include "base/macros.h"
+#include "build/build_config.h"
```
jmarantz, line 9: I am wondering how many of these base/* files are really needed. It might be easier just to copy them now though and prune them out later.

base/logging.h
```diff
@@ -0,0 +1,1065 @@
+#pragma once
+
+#include "glog/logging.h"
+
+namespace logging {
+  constexpr int LOG_INFO = google::GLOG_INFO;
```
jmarantz, line 6: we should see whether there's a better place to pull this from. We may get it transitively from gflags or gtest already, without needing Chromium's version. Not sure.

Also could be a follow-up.

jmarantz (approved):
I can see this was quite a slog. I got about 1/3 of the way through and now I think we should just submit and iterate.

PR #1940, opened 2019-08-22
Bazel-branch: re-enable a few tests

oschaaf:
- We can enable `StaticAssetManagerTest.TestJsXXX` now that the closure
  compiler flow is working
- We can enable `Sha1SignatureTest.*`, the tests themselves are fine.
  But they work because of a change I made to `SHA1Signature::RawSign`,
  the version on master segfaults. Root-cause analysis of that is in
  progress.


ashishk-1 (approved):
Yes, tests are working after closure compiler implementation.

PR #1945, opened 2019-08-29
Grpc RpcHandler: Fix finish before start behavior

oschaaf:
We make a major leap forward in the grpc version we use compared
to what we have on master on this branch.
All grpc related tests continued to pass except RpcHandlerTest.FinishBeforeStart

It looks like behavior changed, and this change proposes a way to handle that and amends
test expectations.


oschaaf:
(For context, without this change the original test code would hang at the first call to `client->Finish()`)

ashishk-1 (approved):
great find..looks good to me

pagespeed/controller/rpc_handler_test.cc
```diff
@@ -559,17 +559,16 @@ TEST_F(RpcHandlerTest, FinishAfterCompletion) {
 
 // Check nothing bad happens (leaks or stalls) if Finish is called before the
 // first call to HandleRequest.
-// XXX(oschaaf): this one still fails.
-TEST_F(RpcHandlerTest, DISABLED_FinishBeforeStart) {
+TEST_F(RpcHandlerTest, FinishBeforeStart) {
   MockRpcHandler* handler = new MockRpcHandler(&service_, queue_.get());
-  EXPECT_CALL(*handler, HandleRequest(_)).Times(0);
+  EXPECT_CALL(*handler, HandleRequest(_)).Times(1);
```
hillsp: This doesn't look right. The RpcHandler subclass isn't expecting a call for this case. If it gets one, the Proto won't be populated so suddenly a whole bunch of other edge cases are introduced.

pagespeed/controller/rpc_handler.h
```diff
@@ -228,7 +228,7 @@ void RpcHandler<AsyncService, RequestT, ResponseT>::ReadDone(RefPtrT ref) {
 template <typename AsyncService, typename RequestT, typename ResponseT>
 bool RpcHandler<AsyncService, RequestT, ResponseT>::Finish(
     const ::grpc::Status& status) {
-  if (state_ == FINISHED) {
+  if (state_ == FINISHED || state_ == INIT) {
```
hillsp: I guess that a change in gRPC is making Finish() hang now if the client isn't properly initialized. Probably the call to Finish() is now happening before something that it used to happen after. However, the early exit for INIT here means you're not setting state_ to FINISHED. That's causing the extra call to HandleRequest(), which you definitely don't want. Returning false for this case is fine, but you still need to set state_

hillsp (changes requested):

pagespeed/controller/rpc_handler.h
oschaaf: Thanks a lot for taking a look! and, yeah, that seems correct, what I observed to happen is that ordering seems to have changed. With the latest grpc versions, `InitDone()` is called after `Finish()`. I backed my changes out in favor of a tweak in `InitDone()`.
With that I can retain the original test expectations (except for the grpc status).

hillsp (changes requested):

oschaaf:
@hillsp thanks, just pushed a change that amends the comments you pointed out, as well as reverted the status back to CANCELLED.

pagespeed/controller/rpc_handler.h
```diff
@@ -174,7 +174,7 @@ void RpcHandler<AsyncService, RequestT, ResponseT>::Start() {
   // "this" without calling HandleError(), because state_ is INIT.
   InitResponder(service_, &ctx_, &responder_, cq_,
                 MakeFunction(this, &RpcHandler::InitDone,
-                             &RpcHandler::CallHandleError, RefPtrT(this)));
+                            &RpcHandler::CallHandleError, RefPtrT(this)));
```
hillsp: I think you accidentally removed some whitespace here?

pagespeed/controller/rpc_handler.h
```diff
@@ -189,13 +189,11 @@ void RpcHandler<AsyncService, RequestT, ResponseT>::InitDone(RefPtrT ref) {
   } else {
     // When state is FINISHED, it shouldn't be possible for anything else to be
```
hillsp: Ugh, sorry - another silly nit: You're not relying on the refcount to force a disconnect - you're explicitly forcing it. I think it would be more correct to say "When state is FINISHED, we were shutdown before we started. Propagate that as CANCELLED to any client."

hillsp (approved):

pagespeed/controller/rpc_handler.h
oschaaf: I backed that out

pagespeed/controller/rpc_handler.h
oschaaf: No worries, thanks for taking a good look at this!!
Amended as suggested.

oschaaf:
@hillsp feedback processed, ready for another round!

hillsp (approved):
LGTM! Long live gRPC ;)

PR #1946, opened 2019-08-30
Update .bazelrc

oschaaf:
- Allows running tsan/asan/msan
  (e.g.: `bazel test -c dbg --config=clang-asan //net/... //pagespeed/...`)
- Tweak for getting the merge of pagespeed_automatic.a to work with clang
  (though it looks like we may not need pagespeed_automatic.a)

The sanitizers detect a bunch of issues. So we may have some work here
to clean that up, ideally we'd have clean sanitizer runs so we can
enforce it stays clean in CI.


ashishk-1 (approved):
looks good

PR #1949, opened 2019-09-03
updating out buffer size

sambhajis-gdb:


oschaaf (approved):
x-posting from #pagespeed on slack:

```
Ashish 2:32 PM
Regarding issue SHA1Signature::RawSign,
We see that hmac.c file is upgraded when we use it in bazel wrt pagespeed master branch(using gyp)
This is the change in HMAC constructor
+#include "../../internal.h"
+
 
 uint8_t *HMAC(const EVP_MD *evp_md, const void *key, size_t key_len,
               const uint8_t *data, size_t data_len, uint8_t *out,
               unsigned int *out_len) {
   HMAC_CTX ctx;
-  static uint8_t static_out_buffer[EVP_MAX_MD_SIZE];
-
-  /* OpenSSL has traditionally supported using a static buffer if |out| is
-   * NULL. We maintain that but don't document it. This behaviour should be
-   * considered to be deprecated. */
-  if (out == NULL) {
-    out = static_out_buffer;
-  }
-
   HMAC_CTX_init(&ctx);
   if (!HMAC_Init_ex(&ctx, key, key_len, evp_md, NULL) ||
       !HMAC_Update(&ctx, data, data_len) ||
@@ -94,6 +87,14 @@
   EVP_MD_CTX_init(&ctx->md_ctx);
 }
So your change of replacing NULL -> (uint8_t*)hmac is correct.
In addition to your changes, we can upgrade hmac char array of size 20.
We can make it 64 as per this line
# define EVP_MAX_MD_SIZE                 64/* longest known is SHA512 */
(edited)
Or we can define array as
unsigned char hmac[EVP_MAX_MD_SIZE];
```

Following that, this change makes sense to me, thank you!

PR #1951, opened 2019-09-05
Update optipng

Lofesa:
Update optipng to catch the latest commit in this library PR [#1](https://github.com/apache/incubator-pagespeed-optipng/commit/1fa893528ab66e208fa8b34b6e2518016bc6b1fb)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,10 +1,10 @@
 # The two httpds are the longest syncs, so we'd like to start them first.
 [submodule "third_party/httpd/src"]
 	path = third_party/httpd/src
-	url = git://git.apache.org/httpd.git
+	url = https://git.apache.org/httpd.git
 [submodule "third_party/httpd24/src"]
 	path = third_party/httpd24/src
-	url = git://git.apache.org/httpd.git
+	url = https://git.apache.org/httpd.git
 [submodule "third_party/brotli/src"]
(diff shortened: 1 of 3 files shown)


oschaaf:
@Lofesa the system test `inline_unauthorized_resources allows inlining` seems to consistently fail with this, any idea why that is?

[1] https://github.com/apache/incubator-pagespeed-mod/blob/b4bf44cc56d8bbf17494c540dfb6ef20dfcf5073/pagespeed/system/system_tests/inline_unauth.sh#L26

oschaaf:
(triggered CI on master, to see if this specific failure happens there too)

Lofesa:
@oschaaf  No, I don´t know why the test fail, seems unrelated, the only change in this PR is update optipng library to the last commit on it. I don´t know if this PR is OK cause not only changes the folder, it add the changed file in the optipng library and includes the changes from the PR #1923. But I leave the PR as a reminder of this task and don´t close it.

EDIT: The test fail with a timeout. The latest updates to the Data-srcset PR fails in timeout, but not in the same test each time.

oschaaf:
The latest on master fails on the same test, so it's probably an external dependency which has changed. I'll have a look at that

oschaaf:
There are no more commits in this PR that aren't already merged to master. Closing!

PR #1955, opened 2019-09-18
Update LICENSE & NOTICE

oschaaf:
Some changes as per discussed on the email thread. 
Initiating this PR for addressing issues that need to be fixed before we can attempt a release.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -176,6 +176,17 @@
 
    END OF TERMS AND CONDITIONS
 
+   APPENDIX: How to apply the Apache License to your work.
+
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
 ~~~ MOD_PAGESPEED SUBCOMPONENTS
 
 License: Expat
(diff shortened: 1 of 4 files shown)


jmarantz (approved):

jmarantz (approved):

LICENSE
```diff
@@ -176,17 +176,7 @@
 
    END OF TERMS AND CONDITIONS
 
-   APPENDIX: How to apply the Apache License to your work.
```
jukka: Please keep this part.

NOTICE
```diff
@@ -1,110 +1,94 @@
...
 
-Files: src/third_party/google-sparsehash/*
-Copyright: 2005 Google Inc.
-License: BSD-3-clause
+Notice utf
 
-Files: src/third_party/libwebp/*
-Copyright: 2010 Google Inc.
-License: BSD-3-clause
+/*
```
jukka: It's sufficient to keep this part just in the LICENSE file.

LICENSE
oschaaf: Whoops, done

NOTICE
oschaaf: removed this here.

oschaaf (commented):
@jukka feedback addressed in dd6e061bb986094a9189f100e384540122f8639b
The changes to the script that produce the tarball may need a follow up if there are dependent
scripts. I also need to test its result, to double check its contents are buildable / pass tests.
These things would bloat this PR, so I'll do that later.

oschaaf:
Pushed a small change to fix up `devel/build_release_tarball.sh` for the adjusted paths.

PR #1956, opened 2019-09-19
Update libpng

Lofesa:
Update libpng to version 1.6.38


oschaaf:
I'm going to close this, as the current state of master already is on the latest & greatest these days,
and we no longer use git submodules to manage that.

PR #1965, opened 2019-11-27
Fix httpd and http24 dependencies

oschaaf:
The commits these dependencies pointed to have gone away after the ASF source of truth
moved from git.apache.org to github.com.

httpd gets updated to cef6805cb18886c5454a38f3501c5e3c990c0b3d (2.2.29)
httpd24 gets updated to 3cdd2d84507d2a86eb5dab2eec96f1035bb76d6c (2.4.26)

These commits should be equivalent to the refs these deps pointed to before the source of truth moved.


oschaaf:
CI fails on a tests against gstatic.com with respect to inlining unauthorized resources. 
The good news is that the deps are pulled in again with this change.

jmarantz (approved):

PR #1969, opened 2019-12-10
Integration of envoy fetcher in pagespeed

ashishk-1:
Added support for envoy fetcher in pagespeed using bazel build.


pagespeed/envoy/cluster.yaml
```diff
@@ -0,0 +1,41 @@
+static_resources:
```
oschaaf: Instead of this file, can we use the api's to set up the cluster?
for an example, see https://github.com/envoyproxy/nighthawk/blob/master/source/client/process_impl.cc#L232

pagespeed/envoy/BUILD
```diff
@@ -116,3 +120,116 @@ envoy_cc_test(
...
+        "pagespeed_remote_data_fetcher.cc",
+        "envoy_fetch.cc",
+        "header_utils.h",    
+    ],
+    hdrs = [
+        ":envoy_fetcher_headers",
+    ],
+    repository = "@envoy",
+    copts = [
+        "-Iexternal/mod_pagespeed/third_party/apr/gen/arch/linux/x64/include/",
```
oschaaf: Why do we need apr?

pagespeed/envoy/envoy_cluster_manager.cc
```diff
@@ -0,0 +1,104 @@
...
+  ssl_context_manager_ =
+      std::make_unique<Envoy::Extensions::TransportSockets::Tls::ContextManagerImpl>(time_system_);
+
+  cluster_manager_factory_ = std::make_unique<Envoy::Upstream::ProdClusterManagerFactory>(
+      admin_, Envoy::Runtime::LoaderSingleton::get(), store_root_, tls_, generator_,
+      dispatcher_->createDnsResolver({}), *ssl_context_manager_, *dispatcher_, *local_info_,
+      secret_manager_, validation_context_, *api_, http_context_, *access_log_manager_,
+      *singleton_manager_);
+
+  Envoy::MessageUtil::loadFromFile(
```
oschaaf: Yeah, I feel we'd rather build up the cluster proto using the api's here.

pagespeed/envoy/envoy_cluster_manager.h
```diff
@@ -0,0 +1,68 @@
+
+#ifndef PAGESPEED_ENVOY_CLUSTER_MANAGER_H_
```
oschaaf: nit: just stick with #pragma

pagespeed/envoy/envoy_cluster_manager.cc
```diff
@@ -0,0 +1,104 @@
+#include "envoy_cluster_manager.h"
```
oschaaf: ASF license banner needs to be added in a couple of places.

pagespeed/envoy/envoy_cluster_manager.cc
```diff
@@ -0,0 +1,104 @@
...
+
+#include <sys/file.h>
```
oschaaf: is this include needed here?

pagespeed/envoy/envoy_cluster_manager.cc
```diff
@@ -0,0 +1,104 @@
...
+  // that are available.
+  Envoy::Logger::Registry::setLogLevel(level);
+  Envoy::Logger::Logger* logger_to_change = Envoy::Logger::Registry::logger("main");
+  logger_to_change->setLevel(level);
+}
+
+void EnvoyClusterManager::initClusterManager() {
+  auto logging_context = std::make_unique<Envoy::Logger::Context>(spdlog::level::from_str("trace"),
+                                                                  "[%T.%f][%t][%L] %v", log_lock);
+  configureComponentLogLevels(spdlog::level::from_str("trace"));
```
oschaaf: do we really want verbose tracing enabled?

pagespeed/envoy/envoy_cluster_manager.cc
```diff
@@ -0,0 +1,104 @@
...
+
+  api_ = std::make_unique<Envoy::Api::Impl>(platform_impl_.threadFactory(), store_root_,
```
oschaaf, line 83: It would be nice to add a custom log sink which would forward this to PageSpeed's message handler concept.

pagespeed/envoy/envoy_cluster_manager.cc
```diff
@@ -0,0 +1,104 @@
...
+          *dispatcher_, tls_, {}, *local_info_, init_manager_, store_root_, generator_,
+          Envoy::ProtobufMessage::getStrictValidationVisitor(), *api_)});
+
+  singleton_manager_ = std::make_unique<Envoy::Singleton::ManagerImpl>(api_->threadFactory());
+  Envoy::Runtime::LoaderSingleton::get();
+
+  ssl_context_manager_ =
+      std::make_unique<Envoy::Extensions::TransportSockets::Tls::ContextManagerImpl>(time_system_);
+
+  cluster_manager_factory_ = std::make_unique<Envoy::Upstream::ProdClusterManagerFactory>(
```
oschaaf, line 104: We may want to look at customizing the cluster manager factory so it instantiates `Extensions::Common::DynamicForwardProxy` for us to materialize support for multiple destinations..

https://github.com/envoyproxy/envoy/blob/master/source/extensions/clusters/dynamic_forward_proxy/cluster.h

pagespeed/envoy/envoy_fetch.cc
```diff
@@ -0,0 +1,151 @@
...
+      content_length_known_(false) {
+}
+
+EnvoyFetch::~EnvoyFetch() {
+}
+
+void EnvoyFetch::FetchWithEnvoy() {
+  envoy::api::v2::core::HttpUri http_uri;
+  http_uri.set_uri("http://localhost:80");
+  http_uri.set_cluster("cluster1");
```
oschaaf: "cluster1"  is probably tied to the code that set ups the cluster. 
define a const static string or some such instead of hard-coding it?

pagespeed/envoy/envoy_fetch.cc
```diff
@@ -0,0 +1,151 @@
...
+      content_length_(-1),
+      content_length_known_(false) {
+}
+
+EnvoyFetch::~EnvoyFetch() {
+}
+
+void EnvoyFetch::FetchWithEnvoy() {
+  envoy::api::v2::core::HttpUri http_uri;
+  http_uri.set_uri("http://localhost:80");
```
oschaaf: Hard coded?

pagespeed/envoy/envoy_fetch.cc
```diff
@@ -0,0 +1,151 @@
...
+  http_uri.set_cluster("cluster1");
+  std::string uriHash("123456789");
```
oschaaf: What's this?

pagespeed/envoy/envoy_fetch.cc
```diff
@@ -0,0 +1,151 @@
...
+  cluster_manager_.getDispatcher()->run(Envoy::Event::Dispatcher::RunType::Block);
+}
+
+
+// This function is called by EnvoyUrlAsyncFetcher::StartFetch.
+void EnvoyFetch::Start() {
+  std::function<void()> fetch_fun_ptr =
+      std::bind(&EnvoyFetch::FetchWithEnvoy, this);
+  cluster_manager_.getDispatcher()->post(fetch_fun_ptr);
+  cluster_manager_.getDispatcher()->run(Envoy::Event::Dispatcher::RunType::NonBlock);
```
oschaaf, line 96: This looks a bit off, I assume we're posting to an up and running event loop, I wonder why we need to call run on that here?

pagespeed/envoy/envoy_fetch.cc
```diff
@@ -0,0 +1,151 @@
...
+  http_uri.set_uri("http://localhost:80");
+  http_uri.set_cluster("cluster1");
+  std::string uriHash("123456789");
+
+  cb_ptr_ = std::make_unique<PagespeedDataFetcherCallback>(this);
+  std::unique_ptr<PagespeedRemoteDataFetcher> PagespeedRemoteDataFetcherPtr = std::make_unique<PagespeedRemoteDataFetcher>(
+    cluster_manager_.getClusterManager(), http_uri, uriHash, *cb_ptr_);
+
+  PagespeedRemoteDataFetcherPtr->fetch();
+  cluster_manager_.getDispatcher()->run(Envoy::Event::Dispatcher::RunType::Block);
```
oschaaf, line 89: This looks a bit weird to me; shouldn't the dispatcher be started early on, and stopped when it is time to tear down the whole module? So it just runs throughout the module's lifetime?

pagespeed/envoy/envoy_fetch.cc
```diff
@@ -0,0 +1,151 @@
...
+  async_fetch_->response_headers()->SetOriginalContentLength(response_body->length());
+  if (async_fetch_->response_headers()->Has(HttpAttributes::kXOriginalContentLength)) {
+    async_fetch_->extra_response_headers()->SetOriginalContentLength(response_body->length());
+  }
+
+  async_fetch_->Write(StringPiece(response_body->toString()), message_handler());
+
+  async_fetch_->Done(true);
+
+  // async_fetch_ = NULL;
```
oschaaf: ?

oschaaf (changes requested):
Thanks! First round of comments

pagespeed/envoy/BUILD
ashishk-1: Removed dependency of apr. Code pushed https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/05e80e98faba31ff5a94f1244b1833fa88a1fae4

pagespeed/envoy/envoy_cluster_manager.h
ashishk-1: Removed #ifndef and kept #pragma only.
Code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/05e80e98faba31ff5a94f1244b1833fa88a1fae4

pagespeed/envoy/envoy_cluster_manager.cc
ashishk-1: added ASF banner at all missing places.
code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/05e80e98faba31ff5a94f1244b1833fa88a1fae4

pagespeed/envoy/envoy_cluster_manager.cc
ashishk-1: yes, it was a mistake. Although we used cluster.yaml file in this file, it was just a string reference. Actual file read operation was getting performed in envoy utility.
Code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/05e80e98faba31ff5a94f1244b1833fa88a1fae4

pagespeed/envoy/envoy_cluster_manager.cc
ashishk-1: I had kept it at trace level to get all dumps since there were many cluster configuration issues earlier.
Now I have moved to error level. Code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/05e80e98faba31ff5a94f1244b1833fa88a1fae4
I will also implement log sink to redirect the logs to pagespeed logging module.

pagespeed/envoy/envoy_cluster_manager.cc
ashishk-1: yes, will implement custom log sink to forward logs to pagespeed message handler.

pagespeed/envoy/envoy_fetch.cc
ashishk-1: Defined cluster1 as static const string. Code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/05e80e98faba31ff5a94f1244b1833fa88a1fae4

As of now, hardcoded cluster from cluster.yaml file is getting referred. We will move this to API based implementation and then will add dynamic proxy support.

pagespeed/envoy/envoy_fetch.cc
ashishk-1: My bad! Removed hardcoding and used actual url value being passed to EnvoyFetch.
Code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/05e80e98faba31ff5a94f1244b1833fa88a1fae4

pagespeed/envoy/envoy_fetch.cc
ashishk-1: This variable name is not correct. Renaming it to **response_content_hash**. This hash value is response content hash (ignored by our PagespeedRemoteDataFetcher class) which just a dummy placeholder needed by envoy. 
Actually we referred RemoteDataFetcher code of envoy and [replicated here](https://github.com/apache/incubator-pagespeed-mod/blob/05e80e98faba31ff5a94f1244b1833fa88a1fae4/pagespeed/envoy/pagespeed_remote_data_fetcher.cc#L24). 
We kept the structure same as original. So passing dummy value.

Code updated here https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/ae9fdd05d8af8d99c3c9f7c31f2e3ef110f8eb04

pagespeed/envoy/envoy_fetch.cc
ashishk-1: Actually I referred serf fetcher code for fetch response. And did this by referring https://github.com/apache/incubator-pagespeed-mod/blob/master/pagespeed/system/serf_url_async_fetcher.cc#L240
But commented it in my code.
Now removed this commented line altogether.

oschaaf: I see; actually setting `async_fetch_` to nullptr is tidy here as we are done with it, and further dereferencing after this will segfault?

pagespeed/envoy/envoy_fetch.cc
oschaaf: Is `const std::string& content_hash` really needed in the constructor of `PagespeedRemoteDataFetcher`? Can't we just remove it?

pagespeed/envoy/envoy_fetch.cc
ashishk-1: yes, dereferencing it after this step causes segfault

pagespeed/envoy/envoy_fetch.cc
ashishk-1: We can remove this. Even earlier we planned to remove it. But I thought it to keep it just for the sake of envoy **RemoteDataFetcher** structure. 
As we are taking RemoteDataFetcher code as is from envoy and updating it just to remove response content hash check. So in future if RemoteDataFetcher structure gets updated, we can easily incorporate that in our code.
Let me know your views.

oschaaf: If `PagespeedRemoteDataFetcher()` can hide the fact that it needs a content hash internally from consumers that seems nice to me; is there any downside to that?
(I'm not sure I understand why it needs to be exposed in the constructor, as that forces this detail upon consumers of `PagespeedRemoteDataFetcher`?).

ashishk-1: We can hide content hash part from consumers. I will make the change and push the code.
There is no downside to it except this is changing constructor signature of original class.
```
RemoteDataFetcher(Upstream::ClusterManager& cm, const ::envoy::api::v2::core::HttpUri& uri,
                    const std::string& content_hash, RemoteDataFetcherCallback& callback);
```

ashishk-1: Code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/26fd5ca9c767bf7fa32a6561c40ceb79572dc6fb

pagespeed/envoy/utility.cc
```diff
@@ -0,0 +1,95 @@
+/*
```
oschaaf: Do we really need to copy utility.cc/h?

pagespeed/envoy/exception.h
```diff
@@ -0,0 +1,56 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
```
oschaaf: I don't think there's any code in mod_pagespeed that throws, do we need this?

pagespeed/envoy/envoy_fetch.h
```diff
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- * 
+ *
```
oschaaf, line 9: Can be back out the reformatting? It makes it harder to figure out what changed.
It's fine to reformat before this lands, but it's better to do so either before or after review.

pagespeed/envoy/envoy_cluster_manager.cc
```diff
@@ -107,13 +109,38 @@ void EnvoyClusterManager::initClusterManager() {
...
-  Envoy::MessageUtil::loadFromFile(
-      "pagespeed/envoy/cluster.yaml",
-      bootstrap, Envoy::ProtobufMessage::getStrictValidationVisitor(), *api_);
-
-  cluster_manager_ = cluster_manager_factory_->clusterManagerFromProto(bootstrap);
+Envoy::Upstream::ClusterManager&
+EnvoyClusterManager::getClusterManager(const GoogleString str_url_) {
+  UriImpl uri(str_url_);
+  try {
+    uri.resolve(*dispatcher_, Envoy::Network::DnsLookupFamily::Auto);
```
oschaaf: We probably don't want to resolve at this point in time?

pagespeed/envoy/envoy_cluster_manager.cc
```diff
@@ -107,13 +109,38 @@ void EnvoyClusterManager::initClusterManager() {
...
+    std::cout.flush();
+  }
+  cluster_manager_ =
+      cluster_manager_factory_->clusterManagerFromProto(createBootstrapConfiguration(uri));
   cluster_manager_->setInitializedCb([this]() -> void { init_manager_.initialize(init_watcher_); });
+  return *cluster_manager_;
+}
+
+const envoy::config::bootstrap::v2::Bootstrap
+EnvoyClusterManager::createBootstrapConfiguration(const Uri& uri) const {
```
oschaaf: I guess the uri / utilities are all there just so you can pass Uri here right?
Let's just remove the argument, or pass in whatever we need for the relevant lines that need it:

```
  socket_address->set_address(uri.address()->ip()->addressAsString());
  socket_address->set_port_value(uri.port());
```

Probably, string address, uint32_t port or something like that suffices?
I think this is going to be temporary anyway, so we might as well just remove the arguments completely, and just hard code it for now until the forward proxy cluster type is introduced.

oschaaf (changes requested):
thanks, another round of feedback

pagespeed/envoy/envoy_cluster_manager.cc
sambhajis-gdb: used hardcoded argument i.e. Host, Port, and Scheme and pass it to createBootstrapConfiguration()
code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/4ab87afdc150b403bcef176c8f4934e8f4ba18d1

pagespeed/envoy/envoy_cluster_manager.cc
sambhajis-gdb: removed in latest code 
code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/4ab87afdc150b403bcef176c8f4934e8f4ba18d1

pagespeed/envoy/exception.h
sambhajis-gdb: remove this file
code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/4ab87afdc150b403bcef176c8f4934e8f4ba18d1

pagespeed/envoy/utility.cc
sambhajis-gdb: remove this file
code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/4ab87afdc150b403bcef176c8f4934e8f4ba18d1

pagespeed/envoy/envoy_cluster_manager.cc
```diff
@@ -107,13 +109,38 @@ void EnvoyClusterManager::initClusterManager() {
...
+}
 
-  Envoy::MessageUtil::loadFromFile(
-      "pagespeed/envoy/cluster.yaml",
-      bootstrap, Envoy::ProtobufMessage::getStrictValidationVisitor(), *api_);
-
-  cluster_manager_ = cluster_manager_factory_->clusterManagerFromProto(bootstrap);
+Envoy::Upstream::ClusterManager&
+EnvoyClusterManager::getClusterManager(const GoogleString str_url_) {
+  UriImpl uri(str_url_);
```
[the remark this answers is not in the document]
sambhajis-gdb: remove UriImpl() in latest code, Used hardcoded value
code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/4ab87afdc150b403bcef176c8f4934e8f4ba18d1

pagespeed/envoy/envoy_fetch.cc
```diff
@@ -0,0 +1,143 @@
...
+static const char cluster_str[] = "cluster1";
+
+PagespeedDataFetcherCallback::PagespeedDataFetcherCallback(EnvoyFetch* fetch) { fetch_ = fetch; }
+
+void PagespeedDataFetcherCallback::onSuccess(Envoy::Http::MessagePtr& response) {
+  fetch_->setResponse(response->headers(), response->body());
+}
+
+void PagespeedDataFetcherCallback::onFailure(FailureReason reason) {
+  std::cout << "PagespeedDataFetcherCallback::onFailure\n";
```
oschaaf: Can we remove the writes to `std::cout` here, and put in a comment about the steps that need to be done here?

pagespeed/envoy/envoy_fetch.cc
```diff
@@ -0,0 +1,143 @@
...
+      fetcher_(NULL),
+      async_fetch_(async_fetch),
+      message_handler_(message_handler),
+      cluster_manager_(cluster_manager),
+      done_(false),
+      content_length_(-1),
+      content_length_known_(false) {
+}
+
+EnvoyFetch::~EnvoyFetch() {
```
oschaaf: Probably can be removed?

pagespeed/envoy/envoy_fetch.cc
```diff
@@ -0,0 +1,143 @@
...
+
+  envoy::api::v2::core::HttpUri http_uri;
+  http_uri.set_uri(str_url_);
+
+  http_uri.set_cluster(cluster_str);
+  cb_ptr_ = std::make_unique<PagespeedDataFetcherCallback>(this);
+  std::unique_ptr<PagespeedRemoteDataFetcher> PagespeedRemoteDataFetcherPtr = 
+      std::make_unique<PagespeedRemoteDataFetcher>(cluster_manager_.getClusterManager(str_url_), http_uri, *cb_ptr_);
+
+  PagespeedRemoteDataFetcherPtr->fetch();
```
oschaaf: nit: This seems to follow a different naming convention.

pagespeed/envoy/envoy_fetch.cc
```diff
@@ -0,0 +1,143 @@
...
+void EnvoyFetch::Start() {
+  std::function<void()> fetch_fun_ptr = std::bind(&EnvoyFetch::FetchWithEnvoy, this);
+  cluster_manager_.getDispatcher()->post(fetch_fun_ptr);
+  cluster_manager_.getDispatcher()->run(Envoy::Event::Dispatcher::RunType::NonBlock);
+}
+
+bool EnvoyFetch::Init() {
+  return true;
+}
+// This function should be called only once. The only argument is sucess or
```
oschaaf: nit: Newline between `}` and `//...` here and below

pagespeed/envoy/envoy_cluster_manager.h
```diff
@@ -0,0 +1,84 @@
...
+#include "external/envoy/source/exe/platform_impl.h"
+#include "external/envoy/source/exe/process_wide.h"
+#include "external/envoy/source/extensions/transport_sockets/tls/context_manager_impl.h"
+#include "external/envoy/source/server/config_validation/admin.h"
+#include "pagespeed/kernel/base/scoped_ptr.h"
+#include "pagespeed/kernel/base/string.h"
+
+namespace net_instaweb {
+
+class EnvoyClusterManager {
```
oschaaf, line 44: Please add some top level documentation about the purpose of this class

pagespeed/envoy/envoy_cluster_manager.h
```diff
@@ -0,0 +1,84 @@
...
+public:
+  EnvoyClusterManager();
```
oschaaf, line 46: Can we have doc comments for the public methods?

pagespeed/envoy/envoy_fetch.h
```diff
@@ -0,0 +1,107 @@
...
+#include "pagespeed/kernel/http/response_headers.h"
+#include "pagespeed/kernel/http/response_headers_parser.h"
+#include "pagespeed_remote_data_fetcher.h"
+
+namespace net_instaweb {
+
+class EnvoyUrlAsyncFetcher;
+class EnvoyFetch;
+
+class PagespeedDataFetcherCallback : public PagespeedRemoteDataFetcherCallback {
```
oschaaf, line 41: Please add some top level doc comments here (as well as for for the public methods).

pagespeed/envoy/envoy_fetch.h
```diff
@@ -0,0 +1,107 @@
...
+    return 0;
+  }
+
+private:
+  // Do the initialized work and start the resolver work.
+  bool Init();
+
+  // Prepare the request and write it to remote server.
+  int InitRequest();
+  // Create the connection with remote server.
```
oschaaf, line 88: Nit: newline before `//...`

pagespeed/envoy/pagespeed_remote_data_fetcher.cc
```diff
@@ -0,0 +1,65 @@
...
+  callback_.onSuccess(response);
+  request_ = nullptr;
+}
+
+void PagespeedRemoteDataFetcher::onFailure(Envoy::Http::AsyncClient::FailureReason reason) {
+  // ENVOY_LOG(debug, "fetch remote data [uri = {}]: network error {}", uri_.uri(), enumToInt(reason));
+  std::cout << "PagespeedRemoteDataFetcher::onFailure \n";
+  std::cout.flush();
+  request_ = nullptr;
+  // callback_.onFailure(FailureReason::Network);
```
oschaaf: What's the plan here?

pagespeed/envoy/pagespeed_remote_data_fetcher.h
```diff
@@ -0,0 +1,95 @@
...
+
+  ~PagespeedRemoteDataFetcher() override;
+
+  // Http::AsyncClient::Callbacks
+  void onSuccess(Envoy::Http::MessagePtr&& response) override;
+  void onFailure(Envoy::Http::AsyncClient::FailureReason reason) override;
+
+  /**
+   * Fetch data from remote.
+   * @param uri remote URI
```
oschaaf: These comments don't seem to match the arguments to `fetch()` below?

pagespeed/envoy/envoy_url_async_fetcher_test.cc
```diff
@@ -0,0 +1,215 @@
...
+    envoy_url_async_fetcher_->Fetch(urls_[idx], &message_handler_, fetches_[idx]);
+  }
+
+  void StartFetches(size_t first, size_t last) {
+    for (size_t idx = first; idx <= last; ++idx) {
+      StartFetch(idx);
+    }
+  }
+
+  std::vector<EnvoyTestFetch*> fetches_;
```
oschaaf, line 196: Shouldn't these be private?

pagespeed/envoy/envoy_url_async_fetcher_test.cc
```diff
@@ -0,0 +1,215 @@
...
+  EnvoyTestFetch* envoy_fetch_;
+  GoogleString test_host_;
+  MockMessageHandler message_handler_;
+};
+
+TEST_F(EnvoyUrlAsyncFetcherTest, FetchURL) {
+  envoy_fetch_ = new EnvoyTestFetch(
+            RequestContext::NewTestRequestContext(thread_system_.get()),
+            mutex_.get());
+  envoy_url_async_fetcher_->Fetch("http://localhost:80", &message_handler_, envoy_fetch_);
```
oschaaf: Hmm, this isn't going to work in CI, right? (I assume there's no server listening there).
Maybe use on of the test urls that Serf uses targetting modpagespeed.com?

oschaaf (changes requested):
thanks! adding few more comments/questions

pagespeed/envoy/envoy_cluster_manager.cc
ashishk-1: Implemented envoy custom log sink https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/98691261e660d67e6f31d0030b47426f5f93f801

pagespeed/envoy/envoy_fetch.cc
ashishk-1: removed std::cout and added TODO tasks.
code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/8a92117ec74a99b4a1dffff479a71ca1b5db7717

pagespeed/envoy/envoy_fetch.cc
ashishk-1: removed as of now. Will add destructor later when needed.
code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/8a92117ec74a99b4a1dffff479a71ca1b5db7717

pagespeed/envoy/envoy_fetch.cc
ashishk-1: updated naming convention of PagespeedRemoteDataFetcherPtr.
code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/8a92117ec74a99b4a1dffff479a71ca1b5db7717

pagespeed/envoy/envoy_cluster_manager.cc
```diff
@@ -0,0 +1,142 @@
...
+#include "external/envoy/source/exe/platform_impl.h"
+#include "external/envoy/source/exe/process_wide.h"
+#include "external/envoy/source/extensions/transport_sockets/tls/context_manager_impl.h"
+#include "external/envoy/source/extensions/transport_sockets/well_known_names.h"
+#include "external/envoy/source/server/config_validation/admin.h"
+#include "external/envoy/source/server/options_impl_platform.h"
+
+namespace net_instaweb {
+
+static const char cluster_str[] = "cluster1";
```
oschaaf: Can we use `constexpr char xxx[]` instead of `static const char xxx[]` here and in the other instances below?

pagespeed/envoy/envoy_fetch.cc
```diff
@@ -0,0 +1,137 @@
...
+#include "pagespeed/kernel/http/request_headers.h"
+#include "pagespeed/kernel/http/response_headers.h"
+#include "pagespeed/kernel/http/response_headers_parser.h"
+#include "pagespeed_remote_data_fetcher.h"
+
+namespace net_instaweb {
+
+// Default keepalive 60s.
+const int64 keepalive_timeout_ms = 60000;
+static const char cluster_str[] = "cluster1";
```
oschaaf: Isn't this define above as well? Can we make it so that we have a single source of truth?

pagespeed/envoy/envoy_cluster_manager.cc
```diff
@@ -0,0 +1,142 @@
...
+  cluster_manager_factory_ = std::make_unique<Envoy::Upstream::ProdClusterManagerFactory>(
+      admin_, Envoy::Runtime::LoaderSingleton::get(), store_root_, tls_, generator_,
+      dispatcher_->createDnsResolver({}), *ssl_context_manager_, *dispatcher_, *local_info_,
+      secret_manager_, validation_context_, *api_, http_context_, *access_log_manager_,
+      *singleton_manager_);
+}
+
+Envoy::Upstream::ClusterManager&
+EnvoyClusterManager::getClusterManager(const GoogleString str_url_) {
+  std::string host_name = "127.0.0.1";
```
oschaaf: use const here and below

pagespeed/envoy/envoy_cluster_manager.cc
```diff
@@ -0,0 +1,142 @@
...
+#include "external/envoy/source/server/options_impl_platform.h"
+
+namespace net_instaweb {
+
+static const char cluster_str[] = "cluster1";
+
+EnvoyClusterManager::EnvoyClusterManager()
+    : init_watcher_("envoyfetcher", []() {}), secret_manager_(config_tracker_),
+      validation_context_(false, false), init_manager_("init_manager"),
+      local_info_(new Envoy::LocalInfo::LocalInfoImpl(
```
oschaaf: can we use RAII for `local_info_`, e.g. `std::make_unique<Envoy::LocalInfo::LocalInfoImpl>`

pagespeed/envoy/envoy_cluster_manager.cc
oschaaf: Thanks for doing that! Going forward, it's easier to not further increase scope, it's probably better to work on landing what we have now as this already is quite a large PR. Feel free to punt things to TODO's where needed.

pagespeed/system/system_rewrite_driver_factory.h
```diff
@@ -417,6 +417,10 @@ class SystemRewriteDriverFactory : public RewriteDriverFactory {
   int num_rewrite_threads_;
   int num_expensive_rewrite_threads_;
 
+  // if true, pagespeed will use envoy fetcher
+  bool use_envoy_fetcher_ = false;
+
+
```
oschaaf: nit: superfluous newline

pagespeed/envoy/envoy_url_async_fetcher_test.cc
```diff
@@ -0,0 +1,219 @@
...
+    envoy_url_async_fetcher_->Fetch(urls_[idx], &message_handler_, fetches_[idx]);
+  }
+
+  void StartFetches(size_t first, size_t last) {
+    for (size_t idx = first; idx <= last; ++idx) {
+      StartFetch(idx);
+    }
+  }
+
+  std::vector<GoogleString> content_starts_;
```
oschaaf: these can be private?
also, can any of these can be `const`'ed?

pagespeed/envoy/envoy_fetch.cc
ashishk-1: Removed comment along with the function, as the function is not used in fetcher. I added this when I took functionality from serf fetcher, but not using anymore.
Code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/8a92117ec74a99b4a1dffff479a71ca1b5db7717

oschaaf (commented):
thanks -- another pass

pagespeed/envoy/envoy_cluster_manager.h
ashishk-1: added description about the class.
change pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/8a92117ec74a99b4a1dffff479a71ca1b5db7717

pagespeed/envoy/envoy_cluster_manager.h
ashishk-1: added doc comments for all public methods
change pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/8a92117ec74a99b4a1dffff479a71ca1b5db7717

pagespeed/envoy/envoy_fetch.h
ashishk-1: added doc comments as well as class description
change pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/8a92117ec74a99b4a1dffff479a71ca1b5db7717

pagespeed/envoy/envoy_fetch.h
ashishk-1: added new line.
code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/8a92117ec74a99b4a1dffff479a71ca1b5db7717

pagespeed/envoy/pagespeed_remote_data_fetcher.cc
ashishk-1: Purpose of this function is to check the cause of fetch failure and then redirect it to callback.
I have removed unnecessary std::cout and added TODO task there.
Code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/8a92117ec74a99b4a1dffff479a71ca1b5db7717

pagespeed/envoy/pagespeed_remote_data_fetcher.h
ashishk-1: updated function comments.
code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/8a92117ec74a99b4a1dffff479a71ca1b5db7717

pagespeed/envoy/envoy_url_async_fetcher_test.cc
ashishk-1: yes, moved to private.
code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/8a92117ec74a99b4a1dffff479a71ca1b5db7717

pagespeed/envoy/envoy_url_async_fetcher_test.cc
ashishk-1: test case updated with fetching of url **http://selfsigned.modpagespeed.com/mod_pagespeed_example/index.html**
Code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/8a92117ec74a99b4a1dffff479a71ca1b5db7717

pagespeed/envoy/envoy_logger.h
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+#pragma once
+
+#include "external/envoy/source/common/common/logger.h"
+#include "pagespeed/kernel/base/message_handler.h"
+
+namespace net_instaweb {
+
+static const char logger_str[] = "main";
```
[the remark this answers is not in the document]
sambhajis-gdb: Done.
code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/fda843264b794b36ea58e8c160c586a56579eda9

pagespeed/envoy/envoy_cluster_manager.cc
sambhajis-gdb: Done.
code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/fda843264b794b36ea58e8c160c586a56579eda9

pagespeed/envoy/envoy_fetch.cc
sambhajis-gdb: Done.
made a single source of clusterName code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/fda843264b794b36ea58e8c160c586a56579eda9

pagespeed/envoy/envoy_cluster_manager.cc
sambhajis-gdb: Done
code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/fda843264b794b36ea58e8c160c586a56579eda9

pagespeed/envoy/envoy_cluster_manager.cc
sambhajis-gdb: Done.
code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/fda843264b794b36ea58e8c160c586a56579eda9

pagespeed/system/system_rewrite_driver_factory.h
sambhajis-gdb: Done.
code pushed in https://github.com/apache/incubator-pagespeed-mod/pull/1969/commits/fda843264b794b36ea58e8c160c586a56579eda9

pagespeed/envoy/envoy_fetch.cc
oschaaf: Yeah I think we should just remove this line. `post()` provides a thread-safe way to post work to a dispatcher running on another thread. But when you call run here, that makes the dispatcher run on this very same thread. That does not make sense to me, and I suspect no good will come of it; is this an optimization of sorts?

pagespeed/envoy/envoy_fetch.cc
oschaaf: So we probably need start to run the dispatcher earlier, after starting up, and then pass in the argument that indicates we want it to remain running until `dispatcher->exit()` is called. Then we can remove this line.
Exit should be called right before we shut down the thread. Please avoid using the destructor for that, a separate `shutdown()` call is warranted here.

pagespeed/envoy/envoy_url_async_fetcher_test.cc
ashishk-1: moved to private as of now. Currently many of these variables are not used but taken from serf async fetcher tests. When we will implement dynamic forward proxy, then we will add more test and all these variables will come into picture then.
Code pushed in https://github.com/apache/incubator-pagespeed-mod/commit/fda843264b794b36ea58e8c160c586a56579eda9

PR #1973, opened 2020-01-08
Typo fix: conjuction ==> conjunction

tomayac:
This affects three files, probably copy & paste is the culprit.


oschaaf:
lgtm, thanks!

PR #1982, opened 2020-02-07
Firefox webp

Lofesa:
Add Firefox UA webp capable. Some don´t send the navigation webp header. Backlist those that are not webp capable.
@oschaaf @jmarantz can someone check if this is the rigth way to do?


oschaaf (approved):
@Lofesa I updated this to work with the latest on master.
This looked great to me as it was, but it needed to be merged with the latest.

Also noticed some language that we probably should eliminate from the code
base, and followed up on that.  I will request a review from you, if you're OK I would 
like to merge this.

PR #1987, opened 2020-04-10
Improve beacon data decoding

hallerm:
Critical image beacon data that does not conform to the expected format
should be ignored. Update the JSON handling so that fields that do not
have the expected type cause the critical image data to be ignored.

### DIFF
--- a/net/instaweb/rewriter/critical_images_finder.cc
+++ b/net/instaweb/rewriter/critical_images_finder.cc
@@ -410,33 +410,38 @@ RenderedImages* CriticalImagesFinder::JsonMapToRenderedImagesMap(
     LOG(WARNING) << "Bad Json rendered image dimensions map";
     return NULL;
   }
-  // Put the extracted map into RenderedImages proto data.
-  RenderedImages* rendered_images = new RenderedImages();
-  Json::Value::Members imgs = json_rendered_image_map.getMemberNames();
-  for (int i = 0, n = imgs.size(); i < n; ++i) {
-    const GoogleString& img_src = imgs[i];
-    int original_width = json_rendered_image_map[img_src].get(
-        kOriginalImageJsonWidthKey, 0).asInt();
-    int original_height = json_rendered_image_map[img_src].get(
-        kOriginalImageJsonHeightKey, 0).asInt();
-    int rendered_width = json_rendered_image_map[img_src].get(
-        kRenderedImageJsonWidthKey, 0).asInt();
-    int rendered_height = json_rendered_image_map[img_src].get(
-        kRenderedImageJsonHeightKey, 0).asInt();
-    int original_area = (original_width * original_height);
-    int rendered_area = (rendered_width * rendered_height);
-    // Store renderedWidth and renderedHeight for the image only if
-    // the rendered sizes are lower than the original sizes by at least the
-    // percentage threshold set.
-    if (100 * rendered_area < original_area *
-        options->image_limit_rendered_area_percent()) {
-      RenderedImages_Image* images = rendered_images->add_image();
-      images->set_src(img_src);
-      images->set_rendered_width(rendered_width);
-      images->set_rendered_height(rendered_height);
+  try {
+    // Put the extracted map into RenderedImages proto data.
+    RenderedImages* rendered_images = new RenderedImages();
+    Json::Value::Members imgs = json_rendered_image_map.getMemberNames();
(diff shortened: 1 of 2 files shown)


oschaaf:
Thanks! It would be great to cover this with a test to make sure this stays put, would you be up for adding one?

oschaaf (commented):
I was going to ask if we also would need to consider wrapping the `parse()` call(s) in exception handling... 

However it looks like https://github.com/open-source-parsers/jsoncpp/issues/390 may be worth checking out first. Maybe we can use the `JSON_USE_EXCEPTION` flag to our benefit here.
Possibly we'd need to upgrade the [jsoncpp dependency](https://github.com/apache/incubator-pagespeed-mod/tree/master/third_party/jsoncpp).

Any thoughts?

hallerm:
The updated version only changes which classes exception of exceptions are being thrown. The default is to `#define JSON_USE_EXCEPTION 1`. In the current version it will then throw `std::runtime_error`. In the new version this is changed to `Json::Exception` extends `std::exception`. This is why the catch is for `std::exception&`: so that if JsonCpp is upgraded it does not quietly break.

Regarding tests, I am having difficulty figuring out how to run the tests. The instructions at https://github.com/apache/incubator-pagespeed-mod/wiki/Development are failing a clone of git://git.apache.org/apr.git, apr-util, httpd. I've been searching the makefiles but so far haven't stumbled upon a target to run the tests. Is there a simple target to run the unit tests?

oschaaf:
re: tests

A system test would probably be most convenient, possibly it is prudent to add one in
https://github.com/apache/incubator-pagespeed-mod/blob/master/pagespeed/system/system_tests/critical_image_beaconing.sh

You can run all system tests via `make apache_test` in `devel/`.

Or you can run the release scripts via `install/build_release.sh --verbose --skip_psol `. This will run the full test set, as well as sort out installing the deps (if you don't want all these deps installed, an (ubuntu 16) vm is convenient for that).

oschaaf:
Also, thanks for looking into the `JSON_USE_EXCEPTION` more closely. I didn't expect it to work like that :-) Should we then also wrap the `parse()` calls?
I went out and updated the jsoncpp dep to the latest, this seems to just work. I'll can look into that separately, as it's fair to say that is of scope here.

hallerm:
> Also, thanks for looking into the `JSON_USE_EXCEPTION` more closely. I didn't expect it to work like that :-) Should we then also wrap the `parse()` calls?

Agreed, it should be assumed that it throws exceptions from all of its functions.

I'll push another commit with that and hopefully a test soon.

> I went out and updated the jsoncpp dep to the latest, this seems to just work. I'll can look into that separately, as it's fair to say that is of scope here.

Good, we are thinking alike on this.

hallerm:
The `make apache_test` fails as it tries to become root to install memcached. Is there really no way to run simple unit tests?

hallerm:
Nevermnd. I found that I can just ignore the install error and run `mod_pagespeed/out/Debug/mod_pagespeed_test` directy.

oschaaf:
If you want to add a c++ unit tests for this, you should be able to just run `make pagespeed_automatic_test mod_pagespeed_test`, and then run either of those directly, sure.
You may be able to use `--gtest_filter=Foo.Bar` as an argument to those, to narrow done to just the one(s) you added for convenience.

jmarantz:
I  think the test you want for this is probably in pagespeed_automatic_test rather than mod_pagespeed_test. The latter has system-y things like memcached, the former has algorithmic thinks like critical_image_finder. 'make apache_test' runs both, but it makes sense for you to just focus on the new test.

hallerm:
Thank you both. I expect this to be very helpful.

oschaaf:
FWIW, off topic, but I also pushed out a simple dev docker image in https://github.com/apache/incubator-pagespeed-mod/commit/dfa4344d5c3d8c0490ec24c1a88d05fdc8bd36f6
(see the commit description on how I used it)
Maybe we should attempt to land such a thing at some point

hallerm:
Ok, I added a test. And I updated to put all of the JsonCpp code inside the try. I'm not sure how to induce errors in those other spots without a bit of research, but at least it covers the one that we were getting hit with.

oschaaf (approved):
Lgtm, let’s see if CI is happy as well

oschaaf:
Thanks @hallerm !

jmarantz:
+1 thanks!

PR #1998, opened 2020-07-21
Several modernizations

oschaaf:
Notable changes:

- Envoy revision: 939ab3911818d0afe59cb9353ba06f1adfa0303e
- Drops libicu, nspr, chromium/base, chromium/googleurl, misc code.
- Switches to use c++17
- Switches to bazel 3.3.1
- Compatibility tweaks to our `absl::string_view` derived version of `StringPiece`,
  which are needed to preserve behaviour when it comes to handling of nullptr valued
  `const char*`'s

Notes: 
- copies in `StringFormatV` from chromium mostly verbatim.
- still need to explicitly set -std=c++17 in some BUILD files, to avoid
   build errors. Not sure where that flag gets flipped and why we need
   to repeat ourselves to make it stay put.

-------

*Prerequisites*

- [x] Holds off switching to c++17, error originating seemingly from googleurl.
- [x] Envoy extension error, undefined references during linkage.


net/instaweb/rewriter/dedup_inlined_images_filter_test.cc
```diff
@@ -39,10 +39,10 @@ namespace net_instaweb {
 namespace {
 
 // Filenames of resource files.
-const char kCuppaPngFilename[] = "Cuppa.png";
-const char kPuzzleJpgFilename[] = "Puzzle.jpg";
+constexpr absl::string_view kCuppaPngFilename = "Cuppa.png";
```
jmarantz: Are these changes necessary? In the original code this gets initialized by the compiler. It's not obvious to me what happens in the new system; I"m surprised it compiles actually. When does the string_view ctor get run?

net/instaweb/rewriter/image_combine_filter_test.cc
```diff
@@ -36,12 +36,12 @@ namespace net_instaweb {
 namespace {
 
 // Filenames of resource files.
-const char kBikePngFile[] = "BikeCrashIcn.png";
-const char kCuppaPngFile[] = "Cuppa.png";
-const char kPuzzleJpgFile[] = "Puzzle.jpg";
-const char kChefGifFile[] = "IronChef2.gif";
+constexpr absl::string_view kBikePngFile = "BikeCrashIcn.png";
```
jmarantz: ditto -- is this change necessary?

net/instaweb/rewriter/process_context.cc
```diff
@@ -58,7 +58,7 @@ ProcessContext::ProcessContext()
   // url/url_util.cc lazily initializes its "standard_schemes" table in a
   // thread-unsafe way and so it must be explicitly initialized prior to thread
   // creation, and explicitly terminated after thread quiescence.
-  url::Initialize();
+  //url::Initialize();
```
jmarantz: Why was this commented out? This seems risky, per comment.

net/instaweb/rewriter/dedup_inlined_images_filter_test.cc
oschaaf: These are needed for analysis::StrFormat to get compile time checking of the format string. See https://abseil.io/docs/cpp/guides/format, which mentions it

net/instaweb/rewriter/process_context.cc
oschaaf: The method no longer exists, as we moved googlurl versions. I’ll look into if there’s an alternative we should use; thanks for pointing this out as I needed that reminder.

oschaaf: I think we're good here. I double checked, and I couldn't find an equivalent in the latest version, so I think it is safe to assume this is no longer needed. I cleaned this up, and added some more cleanup + a fiasco fix in https://github.com/apache/incubator-pagespeed-mod/pull/1998/commits/cca0fadb5f25fb46bf2e32ed94eee46ec85d17fe (based on querying code level todo's I left in the last change).

net/instaweb/rewriter/dedup_inlined_images_filter_test.cc
jmarantz: At least in this case, kCuppaPngFilename is not used as a format string; it's just used in StrCat.

If we do use a statically initialized string constant, would it work to declare that as
```
constexpr char kFoo[] = "foo";
```
?

net/instaweb/rewriter/process_context.cc
jmarantz: Reading the comment I left above about thread-unsafe; is it worth checking whether the latest version self-initializes in a thread-safe way?

The other thing I'm wondering about is between unit tests, does it leave threads running that might make things less deterministic. But if we can get tests to pass then I think it's worth committing, and we can always figure out how to revisit if needed.

net/instaweb/rewriter/dedup_inlined_images_filter_test.cc
oschaaf: Done in 01f90d1a1b585f36f38d0d5c9da6525e28183525

net/instaweb/rewriter/process_context.cc
oschaaf: From what I understood there's now a method to "freeze" modifying the scheme registry. If we want to be pedantic we could call that here to ensure no-one attempts to modify this registry scheme later in a potentially thread unsafe manner. (But we're not using that functionality).

oschaaf: (ps: the c++ tests are passing, currently initiating a tsan-enabled run to check if that shows complaints)

jmarantz (approved):

PR #2008, opened 2020-07-24
[WIP] Envoy native fetcher

oschaaf:
This takes @ashishk-1 his work from https://github.com/apache/incubator-pagespeed-mod/pull/1969,
and upgrades it to the equivalent of working with the current state of Envoy.
 
### DIFF
--- a/README.md
+++ b/README.md
@@ -1,20 +1,3 @@
-# This (bazel) branch
-
-This branch is for landing a gyp to bazel port of the build system.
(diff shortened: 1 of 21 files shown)


oschaaf:
I'm going to merge this in it's current form so this will not break as we update the Envoy dependency.
Both the Envoy extension and Envoy native fetcher are WIP, but they are passing the limited testing that they
have in place.

PR #2017, opened 2020-07-28
Fix a bunch of build warnings

oschaaf:
- Deprecation warnings from the new jsoncpp dep api usage
- Thread safety analysis
- Unused variables
- char as array index in base64 init call
- clean up a few deps
- relocate some small stuff like macros
- use absl instead of libfmt for int to string routines
- use sharding where possible in test execution, exclude it in tests that 
   would fail when doing so by explicitly setting the number of shards to 1
- switch re2 to use the stock BUILD, and patch it to simplify & reduce warning spew.

Calling out the changes in redis_cache & rewrite_driver.cc/h
specifically as these seem rigth to me but are non-trivial.


oschaaf:
I pushed some more enhancements and cleanups in here that are trivial; still the 
changes in redis_cache & rewrite_driver files would be good to emphasise during
review. Those are the only changes that are non-trivial / would hurt when they are
incorrect.

oschaaf:
s/rewrite_driver/rewrite_options/. 
- a new lock is taken there in `RewriteOptions::IsEqual` which should be done according to the thread safety analysis. 
- redis: rewrite `EnsureConnectionAndDatabaseSelection` to deduplicate, and corrected thread safety annotations.

oschaaf:
Filed https://github.com/apache/incubator-pagespeed-mod/issues/2025 as a leftover for this

PR #2021, opened 2020-07-30
Update build_ngx_pagespeed_from_source.html

wernight:
Simplify a bit the manual compilation by avoiding temporary files and guessing some filenames.

Note: Currently it still contains a lot of legacy code. I suggest to have the build instruction under source control. This would not only be expected, it'd also avoid having this script remain backward compatible.

See also https://github.com/wernight/docker-alpine-nginx-pagespeed/issues/5

### DIFF
--- a/html/doc/build_ngx_pagespeed_from_source.html
+++ b/html/doc/build_ngx_pagespeed_from_source.html
@@ -121,18 +121,16 @@ First download ngx_pagespeed:
 
 <pre>
 #[check the <a href="release_notes">release notes</a> for the latest version</a>]
-NPS_VERSION=1.13.35.1-beta
+NPS_VERSION=1.13.35.2-stable
 cd
-wget https://github.com/apache/incubator-pagespeed-ngx/archive/v${NPS_VERSION}.zip
-unzip v${NPS_VERSION}.zip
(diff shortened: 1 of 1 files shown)


oschaaf (approved):
Lgtm; could you elaborate on the build instruction under version control idea? Maybe I can help there but I’m not sure I understand

wernight:
It's true that `html/doc/build_ngx_pagespeed_from_source.html` is under source control. However I'd have under source control in https://github.com/apache/incubator-pagespeed-ngx maybe just as an `INSTALL` at the root pointing.

This way it would only need to be valid for the current commit, then the instructions could be simplified a lot:

```
-nps_dir=$(find . -name "*pagespeed-ngx-${NPS_VERSION}" -type d)
+nps_dir=incubator-pagespeed-ngx-${NPS_VERSION}
 cd "$nps_dir"
-NPS_RELEASE_NUMBER=${NPS_VERSION/beta/}
-NPS_RELEASE_NUMBER=${NPS_VERSION/stable/}
-psol_url=https://dl.google.com/dl/page-speed/psol/${NPS_RELEASE_NUMBER}.tar.gz
-[ -e scripts/format_binary_url.sh ] && psol_url=$(scripts/format_binary_url.sh PSOL_BINARY_URL)
+psol_url=$(scripts/format_binary_url.sh PSOL_BINARY_URL)
```

PR #2028, opened 2020-08-05
Move pagespeed/kernel/base test specifics to test/

oschaaf:
Having test-specifics under `test/` is more convenient with bazel.
This is a fist in a series of PRs to make that happen. 
Major churn, but this should effectively be a no-op change.


jmarantz:
Why is it more convenient with bazel to have test files under test/... ?

FWIW in Google we usually have the tests with the code being tested, and I think it was an artifact of pre-bazel Envoy (probably cmake) to put the tests in a separate directory.

oschaaf:
Well, I find it is useful anytime one wants to make an easy (non bazel-based) distinction  between test and production code. 
One instance of that is running builds of just production code during development, e.g. `bazel build //pagespeed/...`, without remembering build target names. 
Personally, I also find it helpful to have this separation, as this shrinks down the number of files that go into some of the existing directories, leaning up the listings to code relevant to production. Having said that, this is just personal preference -- if you feel this isn't worth it / shouldn't be done, I'd be happy to close this.

jmarantz:
RE shrinking the set of files in each directory: ok but we really should have a better directory organization :)

RE compiling all the prod binaries easily: that's a pretty good argument :)

jmarantz (approved):

PR #2033, opened 2020-08-18
Use absl base64 codec in favor of our own

oschaaf:


jmarantz:
One thing worth thinking about briefly is if this change in base64 library results in different encodings for the same data, what will be affected?

I think it would not be terrible but it's worth going through a thought experiment. I suspect what will happen is during the transition there will be a variety of hash mismatches, which will result in serving resources with private,max-age=300 response headers. But things will settle out quickly.

I think outline_javascript and outline_css resources will become invalid and result in 404s and broken pages, but only transiently.


It might be worth at least checking to see whether the change causes a different encoding to occur. One way to do that would be to leave the existing tests in place.

To tell the truth I'm not sure if anyone will notice these issues (the outline_* filters are terrible anyway). But it's worth just thinking about it a little, and being prepared to answer questions.

I *do* think it might break some unit tests though...have you noticed?

oschaaf:
All unit tests succeed on my dev machine.
As far as I can tell, there are no direct unit-tests for the methods involved here. 
But there are a bunch of tests hitting them by proxy (e.g. for the hasher, some image utils),
and these don't seem to mind this change. Hopefully Travis will confirm as well in about half an hour.
 
It would be good to get the system tests up and running for additional functional coverage.

Maybe, upon installing a major version (like what's in progress here), we should completely bust the cache
upon install?

jmarantz:
You can just add an update to https://github.com/apache/incubator-pagespeed-mod/blob/8bffe0eeb3369676442e0acac88739240e34721a/net/instaweb/rewriter/public/rewrite_options.h#L599 in to the PR -- that would probably do it.

oschaaf:
> You can just add an update to
> 
> https://github.com/apache/incubator-pagespeed-mod/blob/8bffe0eeb3369676442e0acac88739240e34721a/net/instaweb/rewriter/public/rewrite_options.h#L599
> 
> in to the PR -- that would probably do it.

Nice, I didn't know that existed!  Added in 2044b08

jmarantz (approved):

PR #2095, opened 2022-08-05
Update to Envoy 5261b42

oschaaf:
- Upgrade to the latest Envoy dependency
- Purge the re2 dependency we manage: we can now rely on Envoy for it
- Changes/cleanup around -std=c++17
- Unbreak SerfUrlAsyncFetcherTest.TestHttpsSucceedsForGoogleCom
- General code-level changes to accommodate the new Envoy version
- Disable the Envoy-native fetcher test: it broke, but it's not used
  we may consider either stripping it or fixing it later.
- Leave a note about RunningOnValgrind needing cleanup and stub it.

Signed-off-by: Otto van der Schaaf <oschaaf@apache.org>


oschaaf:
Just found out that this will need a follow up: the benchmarks broke. 

I forgot that we used to tweak the re2 dependency through a patch, in order for us to use its benchmarking features.
That was dropped as we can piggy back on the re2 dependency that Envoy brings in for all other use-cases, and I only tested the benchmarks just now.
The re2 `//:benchmark` target is invisible in the vanilla dependency, so we need to either wire in a way to keep on applying that patch so we can continue to rely on it, or rewrite our benchmarking tests to not rely on re2's features for that.

jmarantz (approved):

PR #2096, opened 2022-08-06
Get CI functional

oschaaf:
PR to iterate on getting CI functional again. It broke, but also needs some love to work with the new Envoy update:
- update the bazel and clang/llvm versions we use.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ env:
     - BIT_FLAG=
 
 script:
-  - bazel test --test_env=REDIS_PORT=6379 --test_env=MEMCACHED_PORT=11211 -c dbg //test/...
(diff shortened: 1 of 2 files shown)


jmarantz (approved):

oschaaf:
Allright this mostly works now, but `SerfUrlAsyncFetcherTest.TestHttpsSucceedsForGoogleCom` still fails, because the target url we test against started responding in a non-uniform way (CI gets a different response, apparently). Will fix tomorrow, by switching the target uri to a specific uri on modpagespeed.com.

jmarantz:
Or maybe we can pick a different url to test?

oschaaf:
> Or maybe we can pick a different url to test?

Done in https://github.com/apache/incubator-pagespeed-mod/pull/2096/commits/fb15bcbf93fe865b8e16ecebf94c9439fb61c818
Let's see if CI passes with this change. If it does, I intend to merge this.

PR #2105, opened 2022-11-05
Update Envoy 

oschaaf:


oschaaf:
Note: GoogleURL needed to update, because of the GURL api changing, which we pulled in transitively via our Envoy dependency.

oschaaf:
In https://source.chromium.org/chromium/chromium/src/+/8de0ee1ed01cd1444d45263207a7a9c8b4d4463c:url/url_canon_stdstring.h;dlc=ec3532203dd162ad783e3318119cf9d5bd8e3633 `SetQuery` is made private; so we no longer can rely on that.
Instead we now use `SetQueryStr`. Looking  at the change we now pass, but I need to clean up.

jmarantz (approved):
Thanks!

